PR #23, opened 2016-06-07
Updating to 0.3.0

gosseti:
- Use routed components instead of mounting everything at once
- Refactor the `Camera` and `Uploader` components to be controlled by `Capture`
- Add `ActionBar` (navigation and close) that can be shared between all views
- Update `DocumentCapture` instructions overlay
- Update `FaceCapture` instructions overlay
- Handle lifecycle events properly in `Camera`


hjheath:
Looks good to me

PR #35, opened 2016-07-05
Improvement/update to latest core

rfreitas:
Changes were done to the sdk ui in order to conform with the changes done to the onfido-sdk-core.


src/components/Capture/index.js
```diff
@@ -47,19 +50,20 @@ export default class Capture extends Component {
 
     const { actions, socket, documentType } = this.props
     const methods = {
-      'document': (payload) => {
-        payload.isValid = false
-        payload.documentType = documentType
-        socket.sendMessage(JSON.stringify(payload))
-        actions.documentCapture(payload)
+      'document': (methodPayload) => {
+        var data = methodPayload.data;
```
stephencookdev: [minor] You're scattering some semicolons in places. The current code standard for this project is no semicolons

src/components/Confirm/index.js
```diff
@@ -19,7 +19,7 @@ const Previews = (props) =>  {
       <Capture {...props} captures={props.captures} />
       <div className='onfido-actions'>
         <button
-          onClick={() => props.action(props.captures)}
+          onClick={() => {props.action(props.method);}}
```
stephencookdev: You shouldn't need this wrapped in a block, should you?

src/components/Capture/index.js
stephencookdev: There's also no real reason to use `var` in ES6+ - `let` and `const` are more explicit, this should be `const` here

stephencookdev:
👍  LGTM mostly, all of my comments are pretty minor

src/components/Capture/index.js
rfreitas: noted, I'll switch to const

rfreitas: any reason why we are going with no semicolon? there are some cases where no semicolon can cause issues.

src/components/Confirm/index.js
rfreitas: you are right, it was added in testing when I put a console.log inside

rfreitas:
syntax issues have been fixed, so PR will be merged

PR #40, opened 2016-07-25
Release/update npm 0.4.1

rfreitas:
New version for npm:

**Release notes:**
1. Document overlay
2. Steps flow customization
3. Export errors for `require()`/`import` usage

Breaking changes:
None


tpinto:
Beautiful PR. Keep em coming.

package.json
```diff
@@ -1,15 +1,19 @@
...
     "predev": "npm run build",
     "dev": "NODE_ENV=development webpack-dev-server --inline --hot --progress --https",
     "start": "superstatic build -p ${PORT:-8080} --host 0.0.0.0 --gzip -c '{\"rewrites\": [{\"source\":\"**\",\"destination\":\"index.html\"}],\"headers\":[{\"source\":\"**\",\"headers\":[{\"key\":\"Cache-Control\",\"value\":\"max-age=86400\"}]}]}'",
     "prestart": "npm run build",
-    "build": "NODE_ENV=production webpack -p --progress && npm run build:commonjs",
+    "build": "NODE_ENV=production webpack -p --progress && npm run build:commonjs && npm run build:css",
     "build:commonjs": "cross-env BABEL_ENV=commonjs babel src --out-dir lib",
+    "build:css": "cp -R src/style/ lib/style/ && lessc lib/style/style.css lib/style/style.css",
     "prebuild": "mkdir -p dist",
+    "clean": "rimraf lib dist es coverage",
```
stephencookdev: I do worry a bit that adding more and more stuff to this package.json is just going to be something that keeps happening...

Is there any kind of solid plan to minimise the logic stored here?

test/browser/index.js
```diff
@@ -23,25 +23,19 @@ describe('App', () => {
 
   describe('routing', () => {
```
stephencookdev: I'm not really a fan of commenting out tests...

I would definitely vote for creating an issue, with the test source code included in the ticket description, so it can be pulled out of the actual source code entirely.

Commented out code has a nasty habit of lingering way longer than anyone intends for it to

stephencookdev:
My 2 comments are reasonably minor. LGTM for the most part 👍

test/browser/index.js
rfreitas: I'm with you, I'm not a fan of commenting any code. Motivation here was to leave at least one example for when we do implement some tests. (If we never do them, the commented will in fact remain...) But I will try to get a minimal test working, just to serve as an example for when we do start the tests.

package.json
rfreitas: With the latest commit I moved some stuff to webpack. I certainly don't like long scripts inside the package.json, but I do like npm tasks. We can have a little back and forth on the next FrontEnd meeting.

PR #41, opened 2016-07-26
Bug/fix npm require export issue

rfreitas:
A new proper PR.
Which isolates the bug fix, where the NPM module export was not being done properly, failing to export CSS and causing exceptions due to it.


webpack.config.babel.js
```diff
@@ -126,18 +146,37 @@ const config = {
...
+    library: 'Onfido',
+    libraryTarget: 'umd',
+    path: `${__dirname}/lib`,
+    publicPath: '/',
+    target: 'web',
+    filename: 'index.js'
+  },
+  module: {
+    loaders: [
+      ...baseLoaders,
```
stephencookdev: I don't think the `baseLoaders` etc. are necessary are they? You could just do `...config.module.loaders` here, I think?

webpack.config.babel.js
```diff
@@ -126,18 +146,37 @@ const config = {
...
+    ]
+  },
+  plugins: [
+    ...basePlugins
+  ].concat(ENV==='production' ? [new webpack.optimize.OccurenceOrderPlugin()] : []),
+  devServer: undefined
 }
 
-export default config
+export default [config,configNode]
```
stephencookdev: I wonder if `config` and `configNode` are a bit vague for names?

Do you think `distConfig` and `libConfig` are clearer?

stephencookdev:
LGTM 👍

webpack.config.babel.js
rfreitas: Certain loaders from `config` should go into the `configNode`, loaders like the `ExtractTextPlugin`, that's why I put the common ones in a separate object called `baseLoaders`

webpack.config.babel.js
rfreitas: yeah, I guess they are more accurate, but neither truly encapsulates the difference between the two. Is there any name for a bundle like the one in dist? And a name for a bundle which is imported through require and npm like the one in lib? Since those are the actual differences between the two, it would be nice to have a word for each..

PR #48, opened 2016-08-01
Bug/webcam detection

rfreitas:
Added:
1. Proper webcam detection using DetectRTC lib
2. Option to force upload mode on document capture [example here](https://github.com/onfido/onfido-sdk-ui/blob/e53848257a92677101778b873b6bdf0d02d2b86f/src/index.html#L67)

Improved:
1. mobile/desktop detection


src/components/utils/isDesktop.js
```diff
@@ -1,3 +1,3 @@
-const width = (window.matchMedia("(min-width: 800px)").matches)
```
stephencookdev: Might as well get rid of this file entirely, right? If we're gonna use `DetectRTC`, we might as well use it

src/components/Capture/index.js
```diff
@@ -27,6 +30,30 @@ export default class Capture extends Component {
...
+    return this.state.hasWebcam;
+  }
+
+  //Fired when there is an active webcam feed
+  onUserMedia() {
+    this.setState({
+      hasWebcam: true,
+      hasWebcamPermission: true,
+      DetectRTCLoading: false
+    })
```
stephencookdev: Minor comment, but the semicolon usage is pretty inconsistent in these repos atm.

Do you have a backlog task to add linting into the build? The longer we wait, the more painful it will be

src/components/Capture/index.js
```diff
@@ -27,6 +30,30 @@ export default class Capture extends Component {
...
+    DetectRTC.load( _ => {
+      this.setState({
+        DetectRTCLoading: false,
+        hasWebcam: DetectRTC.hasWebcam
+      })
+    });
+  }
+
+  supportsWebcam (){
+    if (DetectRTC.isGetUserMediaSupported && this.state.DetectRTCLoading) return true;
```
stephencookdev: Could this look a little cleaner as something like

```
const supportUnknown = DetectRTC.isGetUserMediaSupported && this.state.DetectRTCLoading;

// err on the side of webcam being supported
return supportUnknown || this.state.hasWebcam
```

Just to make it clearer why you're returning true when `DetectRTCLoading`

src/components/utils/isDesktop.js
rfreitas: It's still used for detecting if it's a mobile device. If it's a mobile we don't want to use the camera even if it exists, because it's too taxing and does not perform well.

src/components/Capture/index.js
rfreitas: I have added lint into the build, but there is no semicolon rule on it. I will put a task in the backlog for adding a rule on semicolons.

src/components/Capture/index.js
rfreitas: Good idea.

src/components/utils/isDesktop.js
stephencookdev: Yes, sorry I wasn't clear. My point was that you're wrapping `DetectRTC` here.

I'm suggesting that you just call `DetectRTC.isMobileDevice` directly, rather than using this `isDesktop` helper to do it.

Or, if you want to wrap `DetectRTC`, then entirely wrap it (so move all references to it, to helpers like this one instead)

rfreitas: I think I understand your point now. I actually removed this file at one point when implementing the changes, but then I realised there were other files using the isDesktop module, so it actually made sense to keep it. Also in the future I might find better code for checking `isDocument`, because DetectRTC's goal is not focused on that feature, and from inspecting the code it actually uses user agents so it might not be so robust.

stephencookdev: Sure, so it sounds like we definitely do want indirection then, if `DetectRTC` as a dependency is unstable

I would recommend using indirection entirely, then. So adding more helpers, like

```
hasWebcam = () => DetectRTC.hasWebcam
```

or whatever

rfreitas: I do have that indirection as well at:

```js
supportsWebcam (){...
```

but since it's only used in this module I did not move it to some place separate.

rfreitas: the missing indirection is the `DetectRTCLoading`

PR #51, opened 2016-08-10
Use new signature to check doc image quality

hjheath:
This should not be merged until the backend has been updated with the new signature.


src/components/Capture/index.js
```diff
@@ -58,7 +58,7 @@ export default class Capture extends Component {
 
   handleMessages = (message) => {
     const { actions } = this.props
-    if (message.is_document) {
```
rfreitas: so the backend has already changed to this signature?
And is the backend still backwards compatible?

PR #52, opened 2016-08-25
Improvement/refactor style to be modular

rfreitas:
## Refactored CSS to be modular

Made all the CSS (actually LESS) style modular. This means that a stylesheet is now particular to a component, and style dependencies need to be declared in code.

## New dependencies on Build step

It uses the [css-modules](https://github.com/css-modules/css-modules) from Webpack to achieve the modularity, by prefixing classes with a rule. The prefixing follows the [pacomo](https://github.com/unicorn-standard/pacomo) standard of `appname-componentname-classname` (some appname and classname actually have `-` in them, which breaks the standard a bit). [See the code change to the build config file here](https://github.com/onfido/onfido-sdk-ui/pull/52/files#diff-3fe40ded64ba5bb89dcae9266f8dc418L25)

## How was the refactoring done?

Before, all the style was global and all of the style was inside of `style/style.css`.
That style has now been moved to separate stylesheets on each component, eg. path: `Component/style.css`. Common style has been moved into a component called Theme, `Theme/style.css`. In order to make use of a stylesheet a component needs to make a direct reference in javascript, eg:

```js
import theme from '../Theme/style.css'

const Complete = ({message, submessage}) => (
  <div>
    <div className={theme.step}>
      <span className={`${theme.icon}  ${style.icon}`}></span>
      <h1 className={`${theme.title} ${theme.center}`}>{message}</h1>
      <p className={`${theme["mbottom-large"]} ${theme.center}`}>{submessage}</p>
    </div>
  </div>
)
```

and the corresponding component style:

```CSS
.icon {
  height: 4rem;
  background-size: contain;
  background-image: url('assets/complete.svg');
  margin-top: 1.5rem;
  margin-bottom: 1.5rem;
}
```

### Interference

The SDK will be integrated on other projects, so you can get CSS interference coming from both sides. Imagine you define a class name `.btn` on the SDK, the webpage where the SDK is integrated might have other external elements with a class `.btn`, so the SDK style might affect the rest of the page or be affected by it. One way to solve is to manually prefix each class, but any manual process is subject to errors, and there were indeed some classes without prefix.

### Easier to maintain

Now that style dependencies are explicitly defined in code, it's easier to reason about unused style, and in fact a lot of style has been removed which was not being used.

### Further manipulation

As an example, since style dependencies are defined in code and prefixing is done automatically, code can be moved more easily and be localized to different projects.


src/components/Camera/index.js
```diff
@@ -76,10 +78,10 @@ export default class Camera extends Component {
     const { method, onUserMedia } = this.props
     return (
       <div>
-        <div className='onfido-video-overlay'>
+        <div className={style["video-overlay"]}>
```
stephencookdev: [minor] these `"` should be `'`, right?

src/components/ActionBar/style.css
```diff
@@ -0,0 +1,15 @@
+.base {
```
stephencookdev: So most all of these new style files are just cutting stuff out from `src/style/style.css` right?

Can you highlight any new style files that actually change anything substantial from `src/style/style.css` and need explicit reviewing?

rfreitas: hopefully none to be honest. The goal was to have no impact on the final style, only on the selectors. The actual standard of the prefixing is probably the point of discussion (which is on the [webpack config](https://github.com/onfido/onfido-sdk-ui/pull/52/files#diff-3fe40ded64ba5bb89dcae9266f8dc418R26)).

Another point to consider is also the way that common style is imported and how it can be overriden. At the moment the order of the import plays a big role and maybe it shouldn't. [See here](https://github.com/onfido/onfido-sdk-ui/pull/52/files#diff-ed0f78e51202e84128626370456b9695R9), `style.icon` overrides `theme.icon` only because of the order of importing above.

src/components/Camera/index.js
rfreitas: dunno to be honest, I always default to `"`, any reason for the `'`?

stephencookdev: Just because of [the style guide](https://github.com/airbnb/javascript/tree/master/react#quotes)

rfreitas: Can't really be spotted by the linter though.. Also the quotes will not be passed as a value to the JSX attribute.
`<Foo style={{ left: "20px" }} />`, the quotes in this example are actually passed as an attribute, however the result of `style["video-overlay"]` is what will be passed as a JSX attribute.

src/components/ActionBar/style.css
stephencookdev: This makes sense.

As we mentioned offline, it might make more sense to call 'themes' instead 'common'.

And then with that, the ordering is a little off, yeah. It would make more sense for any style to override the common style, by default. I guess just ordering 'common' first is fine for now, though

rfreitas:
After another offline discussion with @stephencookdev , we are going with the Theme approach, where the minimal style necessary to customise the look and feel of the SDK can be done by customising the Theme selectors.

Issue #36, opened 2016-07-05
TypeError Exception and Automatic capture of user's face


rfreitas:
Version 0.4.0 85ac7659 (possibly on earlier versions too)

1. `npm run dev`
2. Go to webapp
3. Start flow and press identity
4. Capture document
5. Press Take again
6. Capture document
7. Press confirm
8. Wait for face to automatically captures

Problem:
8. "Uncaught TypeError: Cannot read property 'getScreenshot' of null" fires in the console. Face gets automatically captures without having to press on Take Photo.

rfreitas:
fixed on 0.7.0

PR #57, opened 2016-09-09
Bug/gb 36 exception and unwanted automatic face capture

rfreitas:
Fix for https://github.com/onfido/onfido-sdk-ui/issues/36.
Reason is due to `componentDidMount` and `componentWillUnmount` not being called after the second remount.


stephencookdev:
LGTM :+1:

PR #58, opened 2016-09-12
new spec for the document overlay

rfreitas:
Fix for CX-184.

The size and shape of the overlay is incorrect.

- Aspect ratio should be 1.57
- Width should be ~80% (to be investigated) of camera image width


hjheath:
Looks good to me.
Change names from -per to -percent
Explain commented out equation

Issue #56, opened 2016-09-09
Oncomplete callback is called before flow is completed


rfreitas:
1. Go to document capture step
2. Put a document in front of the camera and wait for its capture
3. Press confirm on the document capture
4. Take a picture of ones face
5. Open console

Problem
4. The `onComplete` is called even thought the confirm button on the face capture hasn't been pressed yet

rfreitas:
Fixed on [0.7.0](https://github.com/onfido/onfido-sdk-ui/releases/tag/0.7.0)

PR #59, opened 2016-09-13
Bug/gb 56 oncomplete called before finished

rfreitas:
Fix for issue: https://github.com/onfido/onfido-sdk-ui/issues/56

Must be reviewed together with: https://github.com/onfido/onfido-sdk-core/pull/18

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "detectrtc": "^1.3.1",
     "lodash": "^4.13.1",
     "object-assign": "^4.1.0",
-    "onfido-sdk-core": "^0.4.0",
+    "onfido-sdk-core": "^0.5.0",
     "preact": "^5.0.0-beta9",
     "preact-compat": "^1.7.1",
     "preact-router": "^2.0.0",
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "test:karma": "karma start test/karma.conf.js --single-run",
     "lint": "eslint {src,test}",
     "link:core": "npm link ../onfido-sdk-core",
-    "unlink:core": "rm -r ./node_modules/onfido-sdk-core && npm install"
+    "unlink:core": "rm -r ./node_modules/onfido-sdk-core && npm install onfido-sdk-core"
   },
   "main": "lib/index.js",
   "keywords": [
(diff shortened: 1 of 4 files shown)


stephencookdev:
What was it in the old code that was actually causing the bug? I can't see it

rfreitas:
Forgot to include the complementary PR: https://github.com/onfido/onfido-sdk-core/pull/18

package.json
```diff
@@ -7,14 +7,17 @@
     "dev": "NODE_ENV=development webpack-dev-server --inline --hot --progress --https",
     "start": "superstatic build -p ${PORT:-8080} --host 0.0.0.0 --gzip -c '{\"rewrites\": [{\"source\":\"**\",\"destination\":\"index.html\"}],\"headers\":[{\"source\":\"**\",\"headers\":[{\"key\":\"Cache-Control\",\"value\":\"max-age=86400\"}]}]}'",
     "prestart": "npm run build",
-    "build": "NODE_ENV=production webpack -p --progress",
-    "prebuild": "npm run clean && mkdir -p dist && mkdir -p lib",
+    "build": "NODE_ENV=production webpack -p --progress ",
+    "build:core": "(cd ../onfido-sdk-core && npm run build)",
```
stephencookdev: This feels really hacky... There must be a better way of doing this?

package.json
```diff
@@ -7,14 +7,17 @@
...
+    "prebuild": "npm run clean && mkdir -p dist && mkdir -p lib && (npm run build:core || true)",
     "clean": "rimraf lib dist",
     "check:src": "npm run lint && npm run test",
     "prepublish": "npm run clean && npm run check:src && npm run build",
     "test": "npm run -s lint && npm run -s test:karma",
     "test:karma": "karma start test/karma.conf.js --single-run",
-    "lint": "eslint {src,test}"
+    "lint": "eslint {src,test}",
+    "link:core": "npm link ../onfido-sdk-core",
```
stephencookdev: Is this really something we should be putting in npm scripts?

`npm link` is pretty standard, and this script makes loads of assumptions about where `onfido-sdk-core` is etc.

package.json
```diff
@@ -7,14 +7,17 @@
...
+    "unlink:core": "rm -r ./node_modules/onfido-sdk-core && npm install"
```
stephencookdev: Probably just `npm install onfido-sdk-core`, if we are to keep this. Would be faster

stephencookdev:
Other than my reasonably minor comments, LGTM 👍

package.json
rfreitas: definitely a total hack, but I needed something at least :/

package.json
rfreitas: what I wanted was simply speed when linking and unlinking the core, but I get the argument, I might drop the npm link one.

package.json
rfreitas: good point

PR #63, opened 2016-09-27
Improvement/upload fallback

rfreitas:
### Added
- Public: The capture screen UI now includes an upload button fallback, for whenever the user is experiencing problems with the webcam.

### Changed
- Internal: Updated to `onfido-sdk-core@0.6.0`, selectors are now more general as in they are no longer specific to each capture type, some new selectors are also being used.
- Internal: `Camera`, `Capture` and `Uploader` have been refactored, the pure part of the components have been separated from the "impure"/state logic part. This adds flexibility and encapsulation.
- Internal: The `Capture` component now orchestrates all the state logic of the `Uploader` component, this allows to join the camera and uploader state logic together, necessary for the upload fallback feature.

Demo: https://soggy-school.surge.sh/

Complementary PR: https://github.com/onfido/onfido-sdk-core/pull/21


rfreitas:
Travis automatic deployment: https://63.pr.onfido-sdk-ui.onfido.surge.sh

src/components/Camera/index.js
```diff
@@ -38,7 +80,7 @@ export default class Camera extends Component {
     if (autoCapture) this.capture.start()
     events.on('onBeforeClose', () => {
       this.capture.stop()
-      route('/', true)
+      //route('/', true)
```
stephencookdev: Accidental comment out?

src/components/Capture/index.js
```diff
@@ -106,54 +109,56 @@ export default class Capture extends Component {
...
     )
   }
 }
 
+const CaptureScreen = ({method, useCapture, hasCaptured, ...other})=> (
+  <div className={classNames({
+    [style.camera]: useCapture && !hasCaptured,
+    [style.uploader]: !useCapture && !hasCaptured
+  })}>
+  {hasCaptured ?
```
stephencookdev: I find this ternary syntax confusing when used like this, quite hard to read

It goes against [the style guide](https://github.com/airbnb/javascript#comparison--nested-ternaries) too, really

I'd rather the logic was moved out somehow

src/components/Confirm/index.js
```diff
@@ -38,32 +38,25 @@ const Previews = ({capture, step, retakeAction, confirmAction} ) =>  {
...
       }
-    } = this.props
+    }) => {
 
-    const captures = {
-      'document': documentCaptures,
-      'face': faceCaptures
-    }[method]
-    const capture = getCapture(captures)
+  const captures = validCaptures[method]
```
stephencookdev: I always prefer going for a naming convention of `x` and `xList` or something, when I have both

Otherwise you have to spot the difference between `x` and `xs`, and it just ends up being confusing

src/components/Uploader/index.js
```diff
@@ -30,64 +38,13 @@ export const UploadError = ({errorMessage}) => (
...
-    return (methods[method] || methods['home'])()
-  }
-}
+export const Uploader = ({method, onImageSelected, uploading, noDocument}) => (
+  <Dropzone
+    onDrop={([ file ])=> onImageSelected(file)}
+    multiple={false}
+    className={style.dropzone}
+  >
+    {uploading && <UploadProcessing /> || <UploadInstructions />}
```
stephencookdev: This logic seems a little unnecessary to inline, too, IMO

rfreitas: yeah, I agree, but I was focusing more on splitting pure code from the state code, but those one liners should at some point be changed to something more readable.

src/components/Confirm/index.js
rfreitas: fair point

src/components/Camera/index.js
rfreitas: not accidental, I was testing what the effect of it was and couldn't really see the difference, but now that I think of it, it might be for the modal logic.

src/components/Capture/index.js
rfreitas: alright, fair point, I will move it out

rfreitas:
In response to the PR split. I agree that it would have been better if it had been split, but the fact is that the refactor was not planned, it came out of necessity. I needed to actually separate some of the logic, because the uploader had to be turned from a state machine with one entry point to a state machine with two entry points (a regular upload vs. an upload started from the webcam). Also I needed to understand what the code was doing and where the line of abstraction and encapsulation was, and separating the logic helped me out on that.

rfreitas:
Travis automatic deployment: https://63.pr.onfido-sdk-ui.onfido.surge.sh

src/components/Capture/index.js
```diff
@@ -106,54 +109,62 @@ export default class Capture extends Component {
...
+    }
+  </div>
+)
+
+const CaptureScreen = ({method, useCapture, hasCaptured, ...other})=> (
+  <div className={classNames({
+    [style.camera]: useCapture && !hasCaptured,
+    [style.uploader]: !useCapture && !hasCaptured
+  })}>
+    {hasCaptured ?
```
rfreitas: @stephencookdev it does indeed look much better now.

rfreitas:
Travis automatic deployment: https://63.pr.onfido-sdk-ui.onfido.surge.sh

Issue #54, opened 2016-09-09
Some life cycle methods do not get called at Camera and Capture, such as `componentWillUnmount `


rfreitas:
1. Go to document capture step
2. Put a document in front of the camera
3. Wait for the confirmation screen to appear
4. If the problem is not visible, press Take again and go to step 1.

Problem:
2. The `screenshot` function of the `Camera` component keeps getting called and it's because the `componentWillUnmount` does not get fired.

rfreitas:
The issue is more general now. Components are not getting their lifecycle methods called, methods such as `componentDidMount` or `componentWillUnmount`.

rfreitas:
This bug has been scoped to the component `Camera` and `Capture`. A fix should solve the issue for these components. A more general bug will be created.

hjheath:
Fixed in https://github.com/onfido/onfido-sdk-ui/pull/71 and https://github.com/onfido/onfido-sdk-ui/pull/181

PR #71, opened 2016-10-06
Fix/component lifecycle issue

rfreitas:
This a fix for the bug #54.

It only solves the lifecycle issues for `Camera` and `Capture`.

The issue is due to `preact` not handling nested pure components well. Sibling and stateful parents often don't get lifecycle method calls if they have pure components associated with them.

The fix found was _impurify_ some components.

This is a hack/workaround and more general fix should be found.

### DIFF
--- a/src/components/Camera/index.js
+++ b/src/components/Camera/index.js
@@ -74,19 +74,26 @@ export default class Camera extends Component {
     }
   }
 
-  //Necessary since componentDidMount is sometimes not called
   webcamMounted () {
     const { autoCapture } = this.props
     if (autoCapture) this.capture.start()
     events.on('onBeforeClose', () => {
+      //TODO remove this once the react modal is implemented
       this.capture.stop()
       route('/', true)
     })
   }
 
-  //Necessary since componentWillUnmount is sometimes not called
   webcamUnmounted () {
     this.capture.stop()
+  }
+
+  componentDidMount () {
+    this.webcamMounted()
+  }
+
+  componentWillUnmount () {
+    this.webcamUnmounted()
   }
 
   screenshot = () => {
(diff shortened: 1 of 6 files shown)


rfreitas:
Travis automatic deployment: https://71.pr.onfido-sdk-ui.onfido.surge.sh

rfreitas:
Travis automatic deployment: https://71.pr.onfido-sdk-ui.onfido.surge.sh

rfreitas:
yeah, it goes missing, that UX still needs some tweaking..

rfreitas:
There is still a bug when the document is retaken a second time, the camera unmount method will not be called.

src/components/Camera/index.js
```diff
@@ -100,12 +107,7 @@ export default class Camera extends Component {
       method, onUserMedia, onUploadFallback,
       faceCaptureClick: this.capture.once,
       countDownRef: (c) => { this.countdown = c },
-      webcamRef: (c) => {
```
rfreitas: this was the hack to figure out when mounting and unmounting the webcam

src/components/Uploader/index.js
```diff
@@ -38,7 +39,10 @@ export const UploadError = ({errorMessage}) => (
   <div className={`${style.text} ${style.error}`}>{errorMessage}</div>
 )
 
-export const Uploader = ({method, onImageSelected, uploading, noDocument}) => (
+//TODO move to react instead of preact, since preact has issues handling pure components
+//IF this component is exported as pure,
+//some components like Camera will not have componentWillUnmount called
```
stephencookdev: I generally find that when putting in comments like these, it's helpful to add a link to the issue tracking the problem

That way you can have more detail, have an issue for tracking, and also have something to grep for easily when the issue is resolved

PR #76, opened 2016-10-10
replaced object assign with es6 syntax

rfreitas:
IE11 does not work:

0002screenshot


rfreitas:
Travis automatic deployment: https://76.pr.onfido-sdk-ui.onfido.surge.sh

PR #83, opened 2016-10-25
fixed surge https certiticate issue

rfreitas:
Surge ssl certificate only works on subdomains of one level eg: `*.surge.sh`, subdomains like `*.*.*.surge.sh` will not work.
Therefore all `.` should be replaced with `-`


rfreitas:
Travis automatic deployment: https://83-pr-onfido-sdk-ui-onfido.surge.sh

PR #101, opened 2016-12-05
made pdf validation consistent across different browsers

rfreitas:
Browsers that render pdfs are no longer allowed to do so. The pdf is sent raw to the server, this is to reduce inconsistencies, since the server rendered version might have a different result from the browser rendered version.


rfreitas:
Travis automatic deployment: https://101-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://101-pr-onfido-sdk-ui-onfido.surge.sh

PR #106, opened 2016-12-07
added blur pre validation to auto capture camera mode

rfreitas:
### Changed
- Internal: added blur/clarity client-side detection when autoCapture is active


rfreitas:
Travis automatic deployment: https://106-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://106-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://106-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://106-pr-onfido-sdk-ui-onfido.surge.sh

TheComfyChair:
I've got a couple of things I'd like to point out, under differing levels of severity. I'm not sure if we have any kind of format for code reviews right now, so I'll just make one up!

__Must fix (Need to be fixed before I +1)__

- Variables defined at the top of function scope is an anti pattern with let/const. As a result there is a lot of unnecessary pollution, especially when counter variables are declared. Variables should be declared in the scope they are required.

__Should fix (but I wont reject the PR if they're not)__

- Don't use arrow functions for functions that don't need to be anonymous (for example, cloneLowResCanvas, timeFunc, lots of others). As these functions are anonymous, they are not easily identifiable in the call stack. 

- Detect blur is complex. Too complex in my opinion. It should be broken up into a couple of individual functions (for example, a function for the iteration through the pixels, one for edge detection). This can be still done in an imperitive way, if desired, but it can also be done in a more functional manner. An example of this is below (note, I threw this together on the train this morning, don't assume it actually works, it's mostly for illustration!). Regardless of how the breaking up is done, doing so will make the functionality a lot better at self documentation. Yes, it's more verbose, but verbosity is good when the functionality is complex enough to have such high levels of nesting! :)

```
function calculateEdgeCount(pixels, edgeIntensityThreshold = 20) {
      const { edgeCount, sumEdgeWidth } = processRows(pixels, edgeIntensityThreshold);
     //Do the other stuff, probably also in another function
 }

function processRows(pixels, edgeIntensityThreshold) {
   pixels.reduce((...args) => processColumn(...args, edgeIntensityThreshold), {
     edgeCount,
     sumEdgeWidth,
   });
}

function processColumn(memo, row, rowIndex, edgeIntensityThreshold) {
  const { edgeCount, sumEdgeWidth } = row.reduce(detectEdge,
    { edgeCount,
      edgeStart: null,
      edgeIntensityThreshold,
      rowIndex,
      previousValue: null,
      sumEdgeWidth: 0,
    });

  return {
    edgeCount: memo.edgeCount + edgeCount,
    sumEdgeWidth: memo.sumEdgeWidth + sumEdgeWidth,
  }
}

function detectEdge(memo, pixelValue, columnIndex) {

  if (pixelValue === 0) {
    return Object.assign({}, memo, {
      edgeStart: columnIndex,
      previousValue: pixelValue,
    });
  }

  if (!memo.edgeStart || columnIndex <= edgeStart || value >= memo.previousValue) {
    return Object.assign({}, memo, {
      previousValue: pixelValue,
    })
  }

  return Object.assign({}, memo, {
    previousValue: pixelValue,
    edgeCount: oldValue >= memo.edgeIntensityThreshold
      ? memo.edgeCount + 1
      : memo.edgeCount,
    sumEdgeWidth: oldValue >= memo.edgeIntensityThreshold
      ? memo.sumEdgeWidth + (columnIndex - memo.edgeStart - 1)
      : memo.edgeCount,
  })

}
```

__Could fix (just small things)__

- Linked to the 'must fix'. Variables should be declared when they are needed. Hoisting doesn't occur with let/const, so there's no advantage to having the declarations at the top of the function, and may actually lead to unnecessary declarations if you return early. 

- All variable declarations should be on a seperate line, without the comma shorthand, in order to make future git commits cleaner. It's also a lot more explicit that you did indeed mean a specific variable to be a const or a let.

- Const should be the default declaration type unless you intend for the assignment to be changed.

src/components/Camera/measure_blur.js
```diff
@@ -0,0 +1,98 @@
+//ref: https://github.com/timotgl/inspector-bokeh/blob/master/measure_blur.js
```
stephencookdev: I'm not sure this is the best source for doing something like this

Personally I would have used something more similar to these methods http://stackoverflow.com/questions/7765810/is-there-a-way-to-detect-if-an-image-is-blurry so the third-party code that you bring in is trying to do something pretty simple and well-defined (e.g. get the laplacian transform, or fast fourier transform)

And the logic that you're bringing into the direct source code is quite minimalist in comparison.

But I don't know - is there a reason you chose this specific "measure blur" source?

rfreitas:
Thanks for the review, Kai. Probably the most comprehensive pr review I have ever seen at Onfido.
This should generate a good discussion ;)

1. Variables defined at the top of the function
So, I don't think I will do changes to `measure_blur.js`, this is copy pasted code, which I don't expect to change without a proper rewrite and not after I actually understand what the algorithm does.
In the rest of the diff I only found one instance where variables are declared outside of their scope, here: https://github.com/onfido/onfido-sdk-ui/pull/106/files#diff-c61c69a9b4227ce0297da4147f4b3f92R118. For this case I will not change it, because I'm forcing `const` values coming in from `props` to be always defined on top, so that I can quickly see which props are being used in the component, since `preact` does not have `propTypes`, this helps a lot.
You mentioned `counter` variables, can you make this specific point on a particular line.
> Variables should be declared in the scope they are required, and the scoping rules/benefits of let/const should be embraced
This point I kinda of agree, but not fully. If you couple declarations to its scope you get more nesting, possible less patterns get revealed and it's harder to parameterise values in the future. Though I do agree there is a balance here!

2. Detect blur is complex.
I agree, but see point 1 response. However, if your code works exactly the same, I will put it in ;)

3. Don't use arrow functions
I disagree with this. Arrow functions are less noisy, they don't have `function` in it and they also make currying very easy and explicit to do. They also mitigate issues with the lexical `this`, even though I agree that class methods that don't need to be binded, don't need to be defined with an arrow function, this way you can get the benefit of prototypal functions. Not fully related to arrow functions, but anonymous functions are great. They make all value types consistent by equating functions to any other value. All other primitives are anonymous, functions should be too. Named functions only come in handy when you need to do recursion, and even that you can work around.

4. Not the biggest fan of hoisting to be honest. Any mathematical solution declares its values before they get used. It's just easier to reason. But I did learn a bit more about hoisting as result of this PR, so thanks for that ;)

5. See your point, but I think there is a place for it. Otherwise you have to extend this to function parameters too. Though I do take the point that some of my lines are way too long.

src/components/Camera/measure_blur.js
rfreitas: because it worked, it was lean and it was packaged as one function that outputted a blur score. All other solutions involved me doing extra bit of work.

stephencookdev: Fair enough.

What are your thoughts on moving it into a `lib/` directory or something, just to make clear that this is 3rd-party code that we're just pulling in?

src/components/utils/math.js
```diff
@@ -0,0 +1,26 @@
+const linearMapping = (intendedStart, intendedEnd, Xstart, Xend, X) =>
+    (X-Xstart)/(Xend-Xstart) * (intendedEnd - intendedStart) + intendedStart
+
+const centeredPos = (sourceWidth, intendedWidth) => (sourceWidth - intendedWidth)/2
```
stephencookdev: These variable names aren't really accurate. The function name confuses me a bit, too.

Does `const getMarginToCenter = (outerLength, innerLength) => (outerLength - innerLength) / 2` make more sense to you?

package.json
```diff
@@ -94,7 +94,9 @@
     "webpack-dev-server": "^1.14.1"
   },
   "dependencies": {
+    "babel-runtime": "^6.20.0",
     "blueimp-load-image": "^2.6.1",
+    "canvasfilters": "git://github.com/kig/canvasfilters.git",
```
stephencookdev: Can't you take the required .js file from here, and put it in the same place as the `measure_blur.js` script (again, a `lib/` folder would make sense to me)

Then you can `dependencies: { "canvasfilters": "file: ./lib/canvasfilters" }`

Just a bit more consistent about how we're pulling in non-npm-managed 3rd-party scripts

rfreitas: sweet 👍

rfreitas:
Travis automatic deployment: https://106-pr-onfido-sdk-ui-onfido.surge.sh

PR #113, opened 2017-01-24
Update expected websockets message signature

hjheath:
The server currently supports both of these signatures. There is not currently a plan or timeline to deprecate the old `is_document` signature.


rfreitas:
Travis automatic deployment: https://113-pr-onfido-sdk-ui-onfido.surge.sh

PR #119, opened 2017-02-15
Feature/cx 553 upload defaults

hjheath:
Document step should default to file upload. Face step should default to webcam when one is available.


rfreitas:
Travis automatic deployment: https://119-pr-onfido-sdk-ui-onfido.surge.sh

src/components/App/StepComponentMap.js
```diff
@@ -11,7 +11,7 @@ const stepToComponents = (stepDefaultOptions, {type: stepType, options: stepOpti
       return [<Select method='document' {...optionExt} />,
               <Capture method='document' {...optionExt} autoCapture={true}/>]
     case 'face':
-      return <Capture method='face' {...optionExt} autoCapture={false}/>
+      return <Capture method='face' useWebcam={true} {...optionExt} autoCapture={false}/>
```
rfreitas: the `useWebcam` can still be customisable correct?

hjheath: Yes, although I agree it's unclear. 

This relies on React reading the last argument if an argument is duplicated. Therefore `<Capture method='face' {...optionExt} useWebcam={true} autoCapture={false}/>` would not be customisable.

I considered writing a function to merge useWebcam in to optionExt if it was not already there, but it felt like overkill.

rfreitas: yeah, I just wanted to confirm if that was the intention. I think we could create two thin wrappers on `Capture` called `FaceCapture` and `DocumentCapture` with different default values to make it a bit more clear. what do you think?

hjheath: Yep that sounds good to me

rfreitas:
Travis automatic deployment: https://119-pr-onfido-sdk-ui-onfido.surge.sh

hjheath:
There's a bug on file upload on the surge deploy, but it works locally.

Do not merge until it's fixed.

rfreitas:
Travis automatic deployment: https://119-pr-onfido-sdk-ui-onfido.surge.sh

src/components/App/StepComponentMap.js
```diff
@@ -33,4 +33,24 @@ const shallowFlattenList = list => [].concat(...list);
...
+  render (options) {
+    return <Capture method='document' autoCapture={true} {...options} />
+  }
+}
+
+DocumentCapture.defaultProps = {
+  useWebcam: false
+}
+
+class FaceCapture extends Component {
```
rfreitas: you can turn this into stateless components, like this:
```
const FaceCapture = options =>
  <Capture method='face' autoCapture={false} {...options} />
```

You can still set `defaultProps` in the same way.

src/components/Capture/index.js
```diff
@@ -222,8 +222,3 @@ const CaptureScreen = ({method, useCapture, hasCaptured, ...other})=> (
     }
   </div>
 )
-
```
rfreitas: not problematic to leave this on, considering that the component can still be used in isolation as it stands now, might be argued that we don't want to do that though.

src/components/utils/file.js
```diff
@@ -1,4 +1,4 @@
-import loadImage from 'blueimp-load-image/js/load-image'
+import loadImage from 'blueimp-load-image/js'
```
rfreitas: are we not using the previous version? something to change to make the import different?

rfreitas:
Travis automatic deployment: https://119-pr-onfido-sdk-ui-onfido.surge.sh

src/components/utils/file.js
hjheath: There is an [index.js](https://github.com/blueimp/JavaScript-Load-Image/blob/master/js/index.js) so node projects can import everything in one statement.

This protects us from future changes and means can use any version.

src/components/Capture/index.js
hjheath: It's not being used, if someone needs a default in the future they can add it.

src/components/App/StepComponentMap.js
hjheath: nice - I will make this change.

src/components/Capture/index.js
rfreitas: then do you think the actually used (wrapped) components should be the ones being exported instead? As it stands now this component does not have any defaults and it will crash if the props are not set. it might be better not to export it at all, and just export the wrappers instead. This is a minor issue, so do as you see best.

src/components/utils/file.js
rfreitas: I see. So the other modules required in the index are also necessary for what we want, I guess?

hjheath: no we only need `load-image` at the current version. If we update to 2.12.0 we would also need `image-scale`. We do not use any of the others.

rfreitas:
Travis automatic deployment: https://119-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Capture/index.js
hjheath: Yeah that makes sense. I am exporting the wrapped components.

rfreitas:
Travis automatic deployment: https://119-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://119-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
btw, what is the build size of `dist/onfido.min.js` on this PR?

hjheath:
266K

rfreitas:
So we got a 10k increase since last release: `https://github.com/onfido/onfido-sdk-ui/blob/0.12.0-rc.1/dist/onfido.min.js`. I wonder if it was because we made the `blueimp-load-image/js/` import more general

README.md
```diff
@@ -204,12 +204,18 @@ A breakdown of the options and methods available to the SDK.
...
-        title:'Open your new bank account'
+      type: 'welcome',
+      options: {
+        title: 'Open your new bank account'
+      }
+    },
+    {
+      type: 'document',
+      options: {
+        useWebcam: true // This is in beta!
```
jcp72: Why make the example use the beta webcam capture feature?

hjheath: No good reason other than to have it documented somewhere. Do you want to remove it?

jcp72: The name of this option is misleading. This affects only the document capture and the use of webcam therein, not face capture. Perhaps _docWebcamCapture_?

hjheath: I disagree, it's nested as part of the document step and so logically refers only to that step.

I think it's simpler to have clear names across the steps. e.g. we use `options` rather than `docOptions`

jcp72: I agree. I read the line out of context.

jcp72: I think it should only be included if it's in addition to a snippet that initiates the SDK with our recommended configuration.

hjheath: The main two examples in the README use the recommended config. This is in the configurations and options section.

PR #129, opened 2017-03-13
Remove stateless components

hjheath:
This fixes bug CX-710.

I fixed this by isolating the commit that caused the bug. I do not have a full understanding of why stateless components caused the camera to break. Let me know if you think we should investigate further.

### DIFF
--- a/src/components/Capture/index.js
+++ b/src/components/Capture/index.js
@@ -14,19 +14,24 @@ import { functionalSwitch, impurify } from '../utils'
 import { canvasToBase64Images } from '../utils/canvas.js'
 import { fileToBase64, isOfFileType, fileToLossyBase64Image } from '../utils/file.js'
 
-export const DocumentCapture = options =>
+const StatelessDocumentCapture = options =>
   <Capture method='document' autoCapture={true} {...options} />
 
-DocumentCapture.defaultProps = {
+StatelessDocumentCapture.defaultProps = {
   useWebcam: false
 }
 
-export const FaceCapture = options =>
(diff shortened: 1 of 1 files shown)


rfreitas:
Travis automatic deployment: https://129-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
I was never able to completely determine the cause, but at the time I concluded that it was probably a bug on preact. But I don't remember if I actually tested replacing preact with react to be sure.

We should probably open a tech debt issue to look into it at some point.

src/components/Capture/index.js
```diff
@@ -14,15 +14,21 @@ import { functionalSwitch, impurify } from '../utils'
 import { canvasToBase64Images } from '../utils/canvas.js'
 import { fileToBase64, isOfFileType, fileToLossyBase64Image } from '../utils/file.js'
 
-export const DocumentCapture = options =>
```
rfreitas: there is an utility called `impurify(statelessComponent)` that will wrap the stateless component into a stateful one. Btw, we should mark this code as a #HACK or #FIXUP, so that we keep track of this workaround.

rfreitas:
Travis automatic deployment: https://129-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://129-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://129-pr-onfido-sdk-ui-onfido.surge.sh

PR #130, opened 2017-03-20
Feature/reverse capture cx 647

hjheath:
The scope it to capture the reverse side of driving licenses and document checks.

This required changes to sdk-core can be seen [here](https://github.com/onfido/onfido-sdk-core/pull/35). 

It should not merged until sdk-core has been released as it will break the latest demo.


rfreitas:
Travis automatic deployment: https://130-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Capture/capture.js
```diff
@@ -0,0 +1,239 @@
+import { h, Component } from 'preact'
```
rfreitas: any reason for renaming this file? It's convention to name it `index`, since it facilitates the import statement.
If this is a compromise for the IDE, I recommend this:  https://atom.io/packages/smart-tab-name

hjheath: It's not IDE. I wrapped the Capture component and put it's public interface i.e. what App consumes in `index.js`. I move the internals to `capture.js` to make the index more readable.

src/index.js
```diff
@@ -62,15 +66,12 @@ function bindEvents (options) {
...
     faceCapture: data => { options.onFaceCapture(stripOneCapture(data)) },
     complete: data => { options.onComplete(strip(data)) }
   }
 
-  forEach(eventListenersMap, (listener, event) => {
-    if (event === 'ready') events.once(event, listener)
-    else events.on(event, listener)
-  })
-
+  forEach(eventListenersMap, (listener, event) => events.once(event, listener))
```
rfreitas: how does this behave if the user goes back and redoes the flow?

src/components/Capture/index.js
```diff
@@ -1,243 +1,36 @@
...
-import style from './style.css'
-import { functionalSwitch, impurify } from '../utils'
-import { canvasToBase64Images } from '../utils/canvas.js'
-import { fileToBase64, isOfFileType, fileToLossyBase64Image } from '../utils/file.js'
+import Capture from './capture.js'
+import { impurify } from '../utils'
 
-const StatelessDocumentCapture = options =>
-  <Capture method='document' autoCapture={true} {...options} />
+const StatelessFrontDocumentCapture = options =>
```
rfreitas: not sure if naming them stateless is necessary

src/components/Document/index.js
```diff
@@ -9,12 +9,10 @@ DocumentNotFound.defaultProps = {
   message: 'We are unable to detect an identity document in this image. Please try again.'
 }
 
-export const DocumentTitle = ({ useCapture, captureTitle, uploadTitle }) =>
-  <div className={theme.title}>{useCapture ? captureTitle : uploadTitle}</div>
-
-DocumentTitle.defaultProps = {
-  captureTitle: 'Place your document in the rectangle',
-  uploadTitle: 'Upload a picture of your document'
+export const DocumentTitle = ({ useCapture, side }) => {
```
rfreitas: having the static text as props allows us to be closer to the possibility of allowing full text customization.

rfreitas: having the static text as props allows us to be closer to the possibility of allowing full text customization.

rfreitas:
Travis automatic deployment: https://130-pr-onfido-sdk-ui-onfido.surge.sh

src/index.js
hjheath: It will not fire again.

I have an assumption throughout that there will only be one confirmed capture for face, document, reverse document. Is that a bad assumption?

src/components/Capture/index.js
hjheath: I agree, but they need a different name to the one we export. I thought 'stateless' was the most descriptive way to differentiate them from the public exported names

src/index.js
rfreitas: it is if the user goes back. I believe the firing multiple times was a fix to a bug related to going back in the flow and redoing the steps.

src/components/Capture/index.js
stephencookdev: I think you could also just

```
export {
  FrontDocumentCapture: impurify(FrontDocumentCapture),
}
```

src/index.js
stephencookdev: ^ sounds important to check!

But this is my only blocking comment in the PR :)

src/components/Capture/index.js
rfreitas: how does import deconstructing work when exporting an hash? Isn't there ambiguity if I export both a hash as default an individual values individually?

rfreitas: @stephencookdev ^ that was a question for you. (wondering if you knew...)

stephencookdev: Ah, sorry - didn't mean to ignore this :)

So as far as I'm aware

```
export {
  a,
  b,
};

export c;

export default {
  d
}
```

just results in what is effectively

```
{
  a,
  b,
  c,
  default: {
    d
  },
}
```

And then `import { a } from ...` would work, `import { d } from ...` would fail, and `import def from ...; def.d` would work

Not sure if that's what you were asking or not?

rfreitas: ugh, so horrible. That's what I expected. So you can't deconstruct a default export if you have non default exports, but can you deconstruct it if you only have one default export?
And is your example really accurate, can I do `import { default } from ...`?

hjheath: @stephencookdev that did not quite work. I had to do:

```
export default {
  FrontDocumentCapture: impurify(FrontDocumentCapture),
}
```

hjheath:
Closing this PR in favour of https://github.com/onfido/onfido-sdk-ui/pull/132

The comments on this PR have all been addressed there

PR #138, opened 2017-05-08
Return File or Blob (File constructor not supported on IE11)

stefaniacardenas:


stefaniacardenas:
Fixed merge conflicts

rfreitas:
Travis automatic deployment: https://138-pr-onfido-sdk-ui-onfido.surge.sh

src/components/utils/file.js
```diff
@@ -1,16 +1,23 @@
 import loadImage from 'blueimp-load-image/js/load-image'
 import {canvasToBase64Images, toLossyImageDataUrl} from './canvas.js'
 
-export const fileToBase64 = (file, callback, errorCallback) => {
+export const convertFile = (file, callback, errorCallback) => {
   const reader = new FileReader();
-  reader.readAsDataURL(file);
-  reader.onload = function () {
-   callback(reader.result)
+  // If PDF return base64. For other images return blob
```
hjheath: Why are PDFs treated differently? Does the client need to handle them differently? If so is it documented?

hjheath:
What's the difference between returning blob and file?

Does the client need to process a blob before  uploading it to the Onfido API? Can you share a snippet of what the upload to the API code looks like in each case?

src/components/utils/file.js
stefaniacardenas: We are converting PDFs to base64 in order to send them to the server through websockets as the implementation only accepts a string. We won't need to do this conversion when we will use the HTTP protocol and PDFs will be treated just as files. 
We are returning a Blob for every capture, being it a PDF, a File or a screenshot from the webcam. Blob and File are very similar because File objects inherit from Blob. We couldn't return File because the file constructor is not supported on IE11. So the client will be handling all the captures in the same way

rfreitas:
@hjheath Sending a blob or a file is the same when using the browser apis.

src/components/Confirm/index.js
```diff
@@ -22,11 +22,11 @@ const FileViewer = ({file:{preview, type}}) =>
...
 
-const Capture = ({capture:{file}}) =>
+const Capture = ({capture:{image}}) =>
   <div className={style.captures}>
-    {isOfFileType(['pdf'], file) ?
-      <FileViewer file={file}/> :
-      <img src={file.preview} className={style.image} />
+    {isOfFileType(['pdf'], image) ?
+      <FileViewer file={image}/> :
+      <img src={URL.createObjectURL(image)} className={style.image} />
```
rfreitas: how heavy is `createObjectURL`? Should this value not be cached? Can it be added to the `Blob` object?

src/components/utils/file.js
rfreitas: I still don't understand why this function returns Blobs sometimes and base64 string other times? Won't this be consumed by the websockets? Doesn't it need to always be base64?

rfreitas: This function needs to be split. One function converts files to base64, another one converts files to Blobs.
On the last point, I would challenge the need of converting files to Blobs, it might not be necessary, considering they are used so interchangeably.

stefaniacardenas:
I've replied in the comment above. Here is a snippet to send a blob to the Onfido API. Blob just like File need to be posted using form data.
```
const forEach = (object, callback) => {
  Object.keys(object).forEach((key) =>{
    callback(object[key], key, object)
  })
}

const objectToFormData = (object) => {
  const formData = new FormData()
  forEach(object, (value, key) => formData.append(key, value))
  return formData
}

const postFormData = ({endpoint, ...bodyOptions}, callback) => {
  const request = new XMLHttpRequest()
  request.open('POST', endpoint, true)

  request.onload = () => {
    if (request.readyState === request.DONE) {
      if (request.status >= 200 && request.status < 400) {
        console.log(request.responseText)
      } else {
        console.log(request.error)
      }
    }
  }
  request.send(objectToFormData(bodyOptions))
}

const sendFile = (capture, applicant_id, endpoint, callback) => {
  const options = {
    file: capture.image, // the Blob returned by capture.image
    applicant_id,
    endpoint, // ie `/v1/applicants/{applicant_id}/documents`
    type: capture.documentType
  }
  postFormData(options, callback)
}

File: src/components/utils/file.js (hunk 1 above)
Comment (rfreitas) [inline src/components/utils/file.js]: it seem `File` inherits from `Blob`, we can return a File or a Blob depending when it's more convenient and we will still respect the `Blob` interface.

**Comment by rfreitas** [comment]
Travis automatic deployment: https://138-pr-onfido-sdk-ui-onfido.surge.sh

**Comment by rfreitas** [comment]
Travis automatic deployment: https://138-pr-onfido-sdk-ui-onfido.surge.sh

**Comment by rfreitas** [comment]
Travis automatic deployment: https://138-pr-onfido-sdk-ui-onfido.surge.sh

File: src/components/utils/file.js [hunk 3]
```diff
@@ -1,14 +1,25 @@
 import loadImage from 'blueimp-load-image/js/load-image'
 import {canvasToBase64Images, toLossyImageDataUrl} from './canvas.js'
 
-export const fileToBase64 = (file, callback, errorCallback) => {
-  const reader = new FileReader();
-  reader.readAsDataURL(file);
-  reader.onload = function () {
-   callback(reader.result)
+export const pdfToBase64 = (file, callback, errorCallback) => {
```
rfreitas: `fileToBase64` is a better name, since any file can be converted to base64

src/components/Capture/capture.js
```diff
@@ -150,20 +150,20 @@ class Capture extends Component {
...
     }
 
     if (isOfFileType(['jpg','jpeg','png'], file)){
       //avoid rendering pdfs or other formats to image,
       //due to inconsistencies between different browsers and the back end
-      fileToLossyBase64Image(undefined, file,
-        lossyBase64 => this.handleFile(lossyBase64, file),
-        error => this.handleFile(undefined, file)
+      fileToLossyBase64Image(undefined, file, (lossyBase64) => {
+        fileToBlob(file, (blob) => this.handleImage(lossyBase64, blob),this.onFileGeneralError)
```
rfreitas: I don't think we need to convert `File` to `Blob`, since `File` is inheriting from `Blob`. what do you think? Also, @stephencookdev do you have an opinion on this?

src/components/utils/file.js
```diff
@@ -23,13 +34,15 @@ const decodeBase64 = (image) => {
   return {integerArray, mimeString}
 }
 
-export const base64toFile = (image) => {
+const createBlob = (arr, type) => {
+  const blob = new Blob([arr], {type})
+  blob.objectUrl = URL.createObjectURL(blob)
```
rfreitas: I have just been reading about this `createObjectURL`, it might not be the best solution. Since it's easy to create a memory leak with this. Your previous solution was actually better, since the ObjectURL got created at presentation time, but it still wasn't taking into account the case where the URL should be deallocated.

rfreitas: @stephencookdev got an opinion on this?

src/components/Capture/capture.js
```diff
@@ -150,20 +150,20 @@ class Capture extends Component {
...
+      },
+        error => this.handleImage(undefined, file)
       )
+
     } else {
```
rfreitas: this code is unreachable I believe

rfreitas:
Travis automatic deployment: https://138-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Confirm/index.js
```diff
@@ -22,11 +22,11 @@ const FileViewer = ({file:{preview, type}}) =>
   </object>
 
 
-const Capture = ({capture:{file}}) =>
+const Capture = ({capture:{image}, screenshotPreview}) =>
```
rfreitas: you can use `base64` instead of `screenshotPreview`, keeps it simpler.

src/components/Confirm/index.js
```diff
@@ -22,11 +22,11 @@ const FileViewer = ({file:{preview, type}}) =>
   </object>
 
 
-const Capture = ({capture:{file}}) =>
+const Capture = ({capture:{image}, base64}) =>
```
rfreitas: just put `base64` inside of `{image, base64}`

src/components/Confirm/index.js
```diff
@@ -35,7 +35,7 @@ const Previews = ({capture, retakeAction, confirmAction} ) =>
   <div className={`${theme.previews} ${theme.step}`}>
     <h1 className={theme.title}>Confirm capture</h1>
     <p>Please confirm that you are happy with this photo.</p>
-    <Capture capture={capture} />
+    <Capture capture={capture} base64={capture.base64}/>
```
rfreitas: `base64` is inside of capture already, no need to split it here.

rfreitas:
Travis automatic deployment: https://138-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://138-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://138-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://138-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://138-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://138-pr-onfido-sdk-ui-onfido.surge.sh

PR #139, opened 2017-05-10
Use indexOf instead of includes for IE compatibility

hjheath:


rfreitas:
Travis automatic deployment: https://139-pr-onfido-sdk-ui-onfido.surge.sh

TheComfyChair:
I always forget .includes was also an ES2015 thing until IE explodes.

PR #141, opened 2017-05-11
Bug/memory leak fix some refactoring

rfreitas:
## The problem

`react-dropzone` was creating `ObjectURL`s but it did not manage them. This created a memory leak since we were not revoking them.

## Technical Implementation

The `ObjectURL` is automatically revoked and removed from the `File` object, whenever `react-dropzone` creates it.

The `ObjectURL` is still used to create lossy `Base64` and to display Blob or Files in the confirmation step. However these `ObjectURL` are created only when needed and are removed (revoked) when no longer necessary.

It's a good practice not to return these `ObjectURL` to the user, since he might not be aware that he needs to revoke them too.

https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL

Internally `image` was renamed `Blob` since it can always be considered a `Blob` even when it's a `File` (since it's inherited).

`imageLossy` was renamed to `lossyBase64` since it's more descriptive.

## Api changes

the returned capture had it's `image` renamed to `blob`. Open to discussion though.

## Documentation Change

Added an example taken from IDV demo app on how to upload captures to webserver

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,13 +241,13 @@ You should get something returned back that looks like this:
     //useful only if you want to track and reference captures which have been taken
     id: "mwxm5loxdu63zlkawcdi",
     //captures will be returned as File or Blob
-    image: File
+    blob: BlobOrFile
     //the result of the document selection step
     documentType: "passport"
   },
   faceCapture: {
     id: "yy5j8hxlxukufjbrzfr",
-    image: Blob,
+    blob: BlobOrFile,
   }
 }
 ```
--- a/README.md
+++ b/README.md
@@ -263,17 +263,49 @@ Onfido.init({
   containerId: 'onfido-mount',
   // here we send the data in the complete callback
   onComplete: function() {
-    var data = Onfido.getCaptures()
-    sendToServer(data)
+    var captures = Onfido.getCaptures()
+    sendToServer(captures)
   }
 })
 
-function sendToServer(data) {
-  var request = new XMLHttpRequest()
-  request.open('POST', '/your/server/endpoint', true)
-  request.setRequestHeader('Content-Type', 'application/json')
-  var dataString = JSON.stringify(data)
-  request.send(dataString)
+const reduceObj = (object, callback, initialValue) =>
+  Object.keys(object).reduce(
+    (accumulator, key) => callback(accumulator, object[key], key, object),
+    initialValue)
+
+const objectToFormData = (object) =>
+  reduceObj(object, (formData, value, key) => {
+    formData.append(key, value)
+    return formData;
+  }, new FormData())
+
+const postData = (data, endpoint, callback) => {
+  const request = new XMLHttpRequest()
+  request.open('POST', endpoint, true)
+
+  request.onload = () => {
+    if (request.readyState === request.DONE) {
+      callback(request)
+    }
+  }
+  request.send(objectToFormData(data))
+}
+
+const sendToServer = ({documentCapture, faceCapture}) => {
+  postFormData({
+    file: documentCapture.blob,
+    type: documentCapture.documentType
+  },
+    '/your-document-endpoint/',
+    request => console.log('document uploaded')
+  )
+
+  postFormData({
+    file: faceCapture.blob
+  },
+    '/your-face-endpoint/',
+    request => console.log('face uploaded')
+  )
 }
 ```
 
--- a/src/components/Capture/capture.js
+++ b/src/components/Capture/capture.js
@@ -100,6 +100,7 @@ class Capture extends Component {
     }
 
     if (!lossyBase64){
+      //If a lossy file cannot be generated, then a raw base64 should be
       fileToBase64(blob, base64 =>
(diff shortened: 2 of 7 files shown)


rfreitas:
Travis automatic deployment: https://141-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
wow, the size is super big, something to do with source maps :x @hjheath your size plugin already saving our ass.

rfreitas:
Travis automatic deployment: https://141-pr-onfido-sdk-ui-onfido.surge.sh

hjheath:
Yes the size plugin is good. I had a look at it would be trivial to change the status of the check if the size change exceeded a given threshold

rfreitas:
Travis automatic deployment: https://141-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
@hjheath that would be neat!

src/components/Capture/capture.js
```diff
@@ -85,29 +85,40 @@ class Capture extends Component {
...
   }
 
-  handleBase64 = (lossyBase64, base64) => {
-    const blob = base64toBlob(base64)
-    this.handleImage(lossyBase64, blob, base64)
-  }
-
-  handleImage = (base64ImageLossy, image, base64) => {
-    if (!image) {
+  handleImage = (lossyBase64, blob) => {
```
hjheath: I think we can use a better name than `handleImage`

src/components/Capture/capture.js
```diff
@@ -85,29 +85,40 @@ class Capture extends Component {
...
       console.warn('Cannot handle a null image')
       return;
     }
-    const payload = this.createPayload(image, base64ImageLossy, base64)
-    functionalSwitch(this.props.method, {
-      document: ()=> this.handleDocument(payload),
-      face: ()=> this.handleFace(payload)
-    })
+
+    const handleImageInternal = (lossyBase64, blob, base64) => {
```
hjheath: The next 20 lines of logic still feels awkward to me. When would we have a lossyBase64 and a base64 argument? If the answer is never then can't this function just have two arguments (blob, base64)?

src/components/Capture/capture.js
rfreitas: suggestions?

webpack.config.babel.js
```diff
@@ -42,6 +42,7 @@ const basePlugins = [
 if (ENV === 'production') {
   basePlugins.push(
     new webpack.optimize.UglifyJsPlugin({
+      sourceMap: true,
```
hjheath: what does this do?

src/components/Capture/capture.js
rfreitas: answer is never, only thing is that to me mixing a lossybase64 and a raw base64 can lead to confusion. They are different base64. Although I agree the whole thing probably needs a rethinking.

README.md
```diff
@@ -263,17 +263,49 @@ Onfido.init({
...
   }
 })
 
-function sendToServer(data) {
-  var request = new XMLHttpRequest()
-  request.open('POST', '/your/server/endpoint', true)
-  request.setRequestHeader('Content-Type', 'application/json')
-  var dataString = JSON.stringify(data)
-  request.send(dataString)
+const reduceObj = (object, callback, initialValue) =>
```
hjheath: These examples look valuable, I would have put them in a separate PR.
1) Do you think they should be part of the SDK code (i.e. we expose `postData` to them).
2) I think we should do the wider rewrite of docs asap.

src/components/Capture/capture.js
rfreitas: thinking more about this and you are right, I'll simplify this.

webpack.config.babel.js
rfreitas: makes sure the minifying plugin does not mess with the source maps. But I still haven't completely solved the issue, I'm still not getting perfect source maps on Chrome.

README.md
rfreitas: fair point on the separate PR.
Making it part of the SDK would not be a bad idea. What do you think @stefaniacardenas?

src/components/Capture/capture.js
hjheath: Ignore, I think `handleImage` is OK. I think `handleImageInternal` can be improved - it does not need to be a function if it's just the implementation of `handleImage`

rfreitas:
Travis automatic deployment: https://141-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://141-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://141-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Capture/capture.js
```diff
@@ -169,14 +156,18 @@ class Capture extends Component {
       return
     }
 
+    const handleFile = file => fileToBase64(file,
```
hjheath: indents are a bit weird, but otherwise looks good!

PR #140, opened 2017-05-10
Fix orientation bug using load-image exif lib

hjheath:
Pull request to see change of size of bundle

PR #143, opened 2017-05-16
Fix so that the confirm component actually uses base64 when it’s a file

rfreitas:
## Problem

The intended was not working correctly. The idea was to use the base64 version of the capture when the capture had been file uploaded. When it had been captured using the webcam, the blob version (highest quality) would be used.
The problem was that the blob was always being used, which meant that the Exit rotation was not being applied.

related PR: https://github.com/onfido/onfido-sdk-ui/pull/140


rfreitas:
Travis automatic deployment: https://143-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
@hjheath do you want me to merge to master or release branch?

PR #145, opened 2017-05-17
Feature/use http cx 754

stefaniacardenas:
Update the JS SDK to use HTTP protocol instead of websockets.
User will see the following error messages in the current UI:
- connection errors (on file upload only)
- validation errors
- file type errors


rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Capture/capture.js
```diff
@@ -23,7 +24,8 @@ class Capture extends Component {
       hasWebcam: DetectRTC.hasWebcam,
       DetectRTCLoading: true,
       uploadFallback: false,
-      fileError: false
+      fileError: false,
```
rfreitas: by looking at the code, it might be simpler to just have one property called `error`, which has both `fileError` and `serverError`.

rfreitas: by looking at the code, it might be simpler to just have one property called `error`, which merges both `fileError` and `serverError`.

src/components/utils/http.js
```diff
@@ -0,0 +1,34 @@
...
+  const url = serverUrl ? serverUrl : SDK_SERVER_URL
+  request.open('POST', `${url}/validate_document`)
+  request.responseType = 'json'
+  request.setRequestHeader('Content-Type', 'application/json')
+  request.setRequestHeader('Authorization', token)
+
+  request.onload = () => {
+    if (request.readyState === request.DONE) {
+      if (request.status === 200) {
+        events.emit('onMessage', request.response)
```
rfreitas: we can bypass the event messaging, by just having a callback passed as parameter. This would remove one dependency from this code and actually make the code easier to read.

rfreitas:
Do we have a plan to remove websocket's code from core as well? pinging @hjheath on this too.

hjheath:
I think we should do the work to merge core and UI. This change makes that simpler.

src/index.js
```diff
@@ -68,11 +48,7 @@ function bindEvents (options) {
     documentCapture: () => options.onDocumentCapture(getCaptures().documentCapture),
     documentBackCapture: () => options.onDocumentCapture(getCaptures().documentBackCapture),
     faceCapture: () => options.onFaceCapture(getCaptures().faceCapture),
-    complete: () => options.onComplete(getCaptures()),
-    onError: () => {
```
rfreitas: I think it might still be important to at least track on Sentry 500 exceptions that happen on the JS client. tracking exceptions that happen due to authorization might be worth it too.

hjheath:
I think you can remove the `Error` component. Worth checking if anything else is redundant now.

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

stefaniacardenas:
We have decided to leave it there in case we need to use it later on for handling other errors, maybe those coming from the webcam.

hjheath:
I disagree with leaving it in. Version control lets us access the old code if we need it. Redundant components add noise

src/components/Router/index.js
```diff
@@ -31,15 +30,16 @@ class Router extends Component {
   }
 
   render = (props) => {
-    const { websocketErrorEncountered, options, ...otherProps} = props
+    const { options, ...otherProps} = props
     const defaultStepOptions = ['welcome','document','face','complete']
     const stepOptions = options.steps || defaultStepOptions
     otherProps.nextStep = this.nextStep
+    otherProps.token = options.token
+    otherProps.serverUrl = options.serverUrl
```
hjheath: What do you think of this @rfreitas? I wrote it, but I'm not thrilled by it.
The options object contains all user config, but we use most of it in parent components. I then explicitly pull out the options we need to pass below the Router. I feel we are halfway between passing all options to all components and only passing the required options.

stefaniacardenas:
That's true @hjheath. I will remove it

src/components/Router/index.js
rfreitas: well. one way to simplify it would be to pass a `postToServer` function with the `token` and `serveurl` binded to it already.

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

src/components/utils/http.js
```diff
@@ -0,0 +1,30 @@
...
+export const postToServer = (payload, serverUrl, token, onComplete) => {
+  const request = new XMLHttpRequest()
+  const url = serverUrl ? serverUrl : SDK_SERVER_URL
+  request.open('POST', `${url}/validate_document`)
+  request.responseType = 'json'
+  request.setRequestHeader('Content-Type', 'application/json')
+  request.setRequestHeader('Authorization', token)
+
+  request.onload = () => {
+    if (request.readyState === request.DONE) {
```
hjheath: I don't think you need this line: http://stackoverflow.com/questions/9181090/is-onload-equal-to-readystate-4-in-xmlhttprequest

src/components/utils/http.js
```diff
@@ -0,0 +1,30 @@
...
+
+export const postToServer = (payload, serverUrl, token, onComplete) => {
+  const request = new XMLHttpRequest()
+  const url = serverUrl ? serverUrl : SDK_SERVER_URL
+  request.open('POST', `${url}/validate_document`)
+  request.responseType = 'json'
+  request.setRequestHeader('Content-Type', 'application/json')
+  request.setRequestHeader('Authorization', token)
+
+  request.onload = () => {
```
hjheath: It might be worth adding and on error callback in case CORS or the server are down: http://stackoverflow.com/questions/10584318/when-should-xmlhttprequests-onerror-handler-fire

src/components/Capture/capture.js
```diff
@@ -80,8 +80,13 @@ class Capture extends Component {
   }
 
   handleMessages = (message) => {
-    const { actions } = this.props
-    const valid = message.valid;
+    const { actions, onError } = this.props
+    if (message.error) {
```
hjheath: I think this error handling should be part of the `onload` callback.

src/components/Capture/capture.js
```diff
@@ -80,8 +80,13 @@ class Capture extends Component {
...
+      this.onServerError()
+      events.emit('onError')
```
hjheath: You could call `Tracker.sendError` directly here rather than using an event here.

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Capture/capture.js
```diff
@@ -123,7 +118,7 @@ class Capture extends Component {
     else {
       postToServer(this.createJSONPayload(payload), serverUrl, token, (response) => {
         this.handleMessages(response)
```
hjheath: you can just pass `this.handleMesssages, this.onServerError`. I would also rename `handleMessages(message)` to `onServerResponse(response)`.

Otherwise looks good!

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Router/index.js
rfreitas: this way you don't have to pass the url and token all the way to this children. do you have other ideas?

src/components/utils/http.js
```diff
@@ -0,0 +1,36 @@
...
+export const postToServer = (payload, serverUrl, token, onSuccess, onError) => {
+  const request = new XMLHttpRequest()
+  const url = serverUrl ? serverUrl : SDK_SERVER_URL
+  request.open('POST', `${url}/validate_document`)
+  request.responseType = 'json'
+  request.setRequestHeader('Content-Type', 'application/json')
+  request.setRequestHeader('Authorization', token)
+
+  request.onload = () => {
+    request.status === 200 ? onSuccess(request.response) : handleError('server error', onError)
```
rfreitas: I think if might be more suitable here. ternary is better understood when there is an assignment involved.

rfreitas: what kind of errors can be found here? 400? 500?

src/index.js
rfreitas: do you guys think it's relevant to have an error callback for the developer? @hjheath @stefaniacardenas

hjheath: Before this change the `onError` callback was not documented and I don't think it ever got fired. I think it is important to let the developer cope with errors - if the server is down or we hit an exception.

Given this change does not remove function I think we need more thought and should deal with it in a separate issue. For example I think the iOS SDK exposes errors in an error object passed to `onComplete`.

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

src/components/utils/http.js
```diff
@@ -13,7 +13,17 @@ const objectToFormData = (object) =>
     return formData;
   }, new FormData())
 
-const handleError = (message, callback) => {
+const errorMessage = (status) => {
+  if (status === 401 || status === 403) return 'unauthorized'
```
hjheath: The server response should include a useful message as well as a status. I think we should log out the message from the server to make it easier for developers to debug

stefaniacardenas:
This message is available on Sentry and will only be visible to us. Developers will see the message from the server response from their chrome tools.

src/index.js
rfreitas: Alright, let's put it in a different issue then. Agree with that.

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://145-pr-onfido-sdk-ui-onfido.surge.sh

src/components/utils/http.js
```diff
@@ -0,0 +1,49 @@
...
+  const url = serverUrl ? serverUrl : SDK_SERVER_URL
+  request.open('POST', `${url}/validate_document`)
+  request.setRequestHeader('Content-Type', 'application/json')
+  request.setRequestHeader('Authorization', token)
+
+  request.onload = () => {
+    if (request.status === 200) {
+      onSuccess(JSON.parse(request.response))}
+    else {
+      handleError(request.status, onError)
```
hjheath: We are still swallowing any error message we get from the server.

src/components/utils/http.js
```diff
@@ -0,0 +1,49 @@
...
+export const postToServer = (payload, serverUrl, token, onSuccess, onError) => {
+  const request = new XMLHttpRequest()
+  const url = serverUrl ? serverUrl : SDK_SERVER_URL
+  request.open('POST', `${url}/validate_document`)
+  request.setRequestHeader('Content-Type', 'application/json')
+  request.setRequestHeader('Authorization', token)
+
+  request.onload = () => {
+    if (request.status === 200) {
+      onSuccess(JSON.parse(request.response))}
```
hjheath: If you set `request.responseType = 'json'` you don't need to parse the response (minor).

hjheath: Just saw it's IE incompatible, maybe add a comment?

rfreitas: definitely a good idea to add a comment

PR #148, opened 2017-05-23
Log error messages on console

stefaniacardenas:
In order to improve the integration experience we print the errors on the console if a requests fails or status code is not 200.

### DIFF
--- a/src/components/utils/http.js
+++ b/src/components/utils/http.js
@@ -22,7 +22,8 @@ const errorMessage = (status) => {
   }
 }
 
-const handleError = (status, callback) => {
+const handleError = ({status, response}, callback) => {
+  console.error(status, response)
   const message = errorMessage(status)
   Tracker.sendError(message)
   callback()
@@ -39,11 +40,11 @@ export const postToServer = (payload, serverUrl, token, onSuccess, onError) => {
     if (request.status === 200) {
       onSuccess(JSON.parse(request.response))}
     else {
-      handleError(request.status, onError)
+      handleError(request, onError)
     }
   }
 
-  request.onerror = () => handleError(request.status, onError)
+  request.onerror = () => handleError(request, onError)
 
   request.send(payload)
 }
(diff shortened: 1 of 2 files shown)


rfreitas:
Travis automatic deployment: https://148-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://148-pr-onfido-sdk-ui-onfido.surge.sh

src/components/utils/http.js
```diff
@@ -22,7 +22,8 @@ const errorMessage = (status) => {
   }
 }
 
-const handleError = (status, callback) => {
+const handleError = ({status, response}, callback) => {
+  console.error(status, response)
   const message = errorMessage(status)
   Tracker.sendError(message)
```
hjheath: Can we send the response message to sentry as well? I think it will be more useful than the generic `errorMessage`

stefaniacardenas: `errorMessage()` is a function that returns a different message based on the status code. The response for 500 errors is "Something went wrong" while for 0 status code, when there is an error in firing the request we don't get a response at all, that's why I'm handling it here.
```const errorMessage = (status) => {
  if (status === 401 || status === 403) return 'unauthorized'
  if (status >= 500) {
    'server error' }
  else {
    'request error'
  }
}```

hjheath: Fine for 500 - although I think sending the status code would be enough - the message does not add much.

But in the case of 401 errors we are still swallowing the reason we denied permission.

You could argue that a 4XX should not be logged to sentry at all, but if we do log them then we should send as much information as possible - i.e. include the response

rfreitas: I agree with Henry, that it's fine to pass the error along. It's one less layer of indirection.
Also, don't worry too much about this logging. We have accomplished the scope of the task which was to convert the protocols, if there is still stuff you guys want to work on it can be added to another issue.

stefaniacardenas: OK, I will do that. Out of curiosity, what type of information do 4xx responses hold?

stefaniacardenas: Oh, I can see it now. It can be related to the referer or the token. Cool, I will add the response body to the Sentry error then. Although for other 4xx errors other than 401 the response body will be empty unless handled on the server. I assume status code will be enough in that case

rfreitas:
Travis automatic deployment: https://148-pr-onfido-sdk-ui-onfido.surge.sh

PR #156, opened 2017-06-21
Feature/pdf preview cx 741

stefaniacardenas:
* Use PDFObject library to display PDFs
* If device is able to preview PDF the confirmation step would show a preview as normal
* If device unable to preview PDF like in IE11 and Edge the confirmation step would show an icon. By clicking on the icon you can download the PDF


rfreitas:
Travis automatic deployment: https://156-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://156-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://156-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://156-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://156-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Confirm/PdfPreview.js
```diff
@@ -0,0 +1,61 @@
...
+    this.embedPDF(previewUrl)
+  }
+  shouldComponentUpdate (nextProps, nextState) {
+    return false;
+  }
+  componentWillReceiveProps({previewUrl}) {
+    if (this.props.pdfPreview !== previewUrl) this.embedPDF(previewUrl)
+  }
+  render() {
+    return <div id='pdfContainer' />;
```
rfreitas: having a non random id makes your component a singleton

stefaniacardenas: Shall I generate the id using css modules, would that work?

stephencookdev: No it wouldn't - CSS modules generate their output based on the file path + file name

Plus it would be a really weird hack to do that :P

Just put a `let i = 0;` outside of the class, and then add a `this.id = 'pdfContainer' + (i++)` to the contructor

stefaniacardenas: OK, done

rfreitas:
Travis automatic deployment: https://156-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://156-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Confirm/PdfPreview.js
```diff
@@ -0,0 +1,67 @@
...
+import PDFObject from 'pdfobject'
+import {preventDefaultOnClick} from '../utils'
+
+const FallbackView = ({blob}) => {
+  const downloadPDF = (e) => {
+    window.navigator.msSaveOrOpenBlob(blob, 'document.pdf')
+  }
+  return (
+    <div>
+      <a href='#' onClick={preventDefaultOnClick(downloadPDF)}><div id='pdfIcon'></div></a>
```
stephencookdev: Can't this be `<a href onClick className={styles.pdfIcon} />`?

Seems silly to have a `<div id='pdfIcon' />` (a singleton inner DOM element) - I can't see that it does anything?

You can just

```
.pdfIcon {
  display: inline-block;
  //...
}
```

in your SCSS

src/components/Confirm/style.css
```diff
@@ -18,9 +18,16 @@
   padding: 0;
 }
 
-.fileViewer{
-  width: 92%;
-  height: 290px;
-  max-height: 70vh;
-  border: 0;
+:global {
+  #pdfIcon {
```
rfreitas: it shouldn't be a global. It should be a class called `.pdfIcon`, which you can then refer to on your JS component like this:
```
<div class=`{style.pdfIcon}`></div>
```

This will namespace your selector, which is great and essential to reduce selector collision with the host app.

src/components/Confirm/style.css
```diff
@@ -18,9 +18,16 @@
   padding: 0;
 }
 
-.fileViewer{
-  width: 92%;
-  height: 290px;
-  max-height: 70vh;
-  border: 0;
+:global {
```
stephencookdev: You should always question what's going on if you're making `:global` stuff for your own code (rather than just for when integrating with third party CSS where it's unavoidable)

In this case, anyway - I believe you don't need to the `:global`, since IDs are just inherently global

But yeah, as per my other comment - this could just be a class of the `a`, I think?

src/components/Confirm/PdfPreview.js
rfreitas: made the same point. Css modules are great!

src/components/Confirm/style.css
stefaniacardenas: Yeah, I know, I was being stupid, I was passing a string when I can just use the embedded html with css modules syntax

src/components/Confirm/style.css
stefaniacardenas: Yes :)

src/components/Confirm/PdfPreview.js
stefaniacardenas: Oh ok, that's why it wasn't displaying. CSS innit

rfreitas:
Travis automatic deployment: https://156-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Confirm/PdfPreview.js
```diff
@@ -0,0 +1,67 @@
+import { h, Component } from 'preact'
+import PDFObject from 'pdfobject'
+import {preventDefaultOnClick} from '../utils'
+import style from './style.css'
+
+const IEPdfBlobLink = ({blob}) => {
```
rfreitas: you might want to explain in a comment why this component is used, as far as I can remember is because IE does not work with generated object URLs on links.

rfreitas:
Travis automatic deployment: https://156-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
I just noticed, this was missing a changelog update

stefaniacardenas:
Done

Issue #153, opened 2017-06-06
`onReady` callback not firing


arnoldmanzano:
How to re-create:
The jsfiddle example posted on the README: https://jsfiddle.net/4xqtt6fL/22/
The `console.log` for onReady does not appear on the logs, the callback doesn't seem to run at any point.
The other events' callback seem to be functioning correctly.

rfreitas:
Hey @arnoldmanzano, it's true. Since we have dropped websockets the `onReady` has lost relevance, since the sdk will boot up and render straight away. Do you still need this functionality in some usecase?

arnoldmanzano:
We are not using the `welcome` step, so it directly loads to the `document` step.
We want to attach some additional `click` `eventListeners` to the buttons on the `document` step (passport, driver's license, identity card buttons). We thought the best way to ensure those buttons are already in the DOM is to attach the listeners after the widget has loaded, using `onReady`.
Since that is not working, we attach the listeners code after `init`, which seems to work, but we cannot confirm if that is a reliable solution.

```
Onfido.init({
  // ...
});

var documentSelectorButtons = document.querySelectorAll('.onfido-sdk-ui-DocumentSelector-option');
  [].forEach.call(documentSelectorButtons, function(button) {
    button.addEventListener('click', function(e) {
    // some code here
  });
});

```

rfreitas:
I see. Well `onReady` would also not give you full certainty that that button is in the DOM, the sdk can always reconstruct, even though it is unlikely.
The most correct way would be to use a live query, but it's also a bit more expensive.

May I ask why you need a listener on the button?

hjheath:
We introduced this bug in `0.15.0` and removed the callback entirely from the SDK in `1.0.0`  https://github.com/onfido/onfido-sdk-ui/releases/tag/1.0.0 so closing this issue as won't fix. 

@arnoldmanzano we are still interested to understand what you use the listener for. Understanding how you use the SDK will help us plan future development.

PR #172, opened 2017-07-06
Feature/upload files cx 849

stefaniacardenas:
In this PR I implemented the file submission from the SDK to Onfido API

Note: JWT v2 implementation is WIP so I have hard-code the applicant_id (not in source code). This will be removed as soon as the JWT implementation is ready

Features:
- Captures (docs and face) are submitted to the Onfido API when the user click 'Confirm' on Confirm screen
- On document upload the is_document validation is handled by the API
- On webcam capture the is_document validation is already handled by the SDK Backend, therefore the advanced validation on API is switched off. 
- Server errors and api authentication errors will be tracked
- I have refactored the way we display the errors on the upload screen in order to reuse error messages across different components
- Updated the documentation


rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

README.md
```diff
@@ -37,22 +37,35 @@ Face capture uses the webcam by default for capturing live photos of users. File
...
+```shell
+$ curl https://api.onfido.com/v2/applicants \
+    -H 'Authorization: Token token=YOUR_API_TOKEN' \
+    -d 'first_name=John' \
+    -d 'last_name=Smith'
+```
+
+You will receive a response containing the applicant id which will be used to create a JSON Web Token.
+
+### 3. Generating a JSON Web Token
```
hjheath: I am calling them SDK Tokens. I think we should be consistent, but I am not sure which is the better name.

README.md
```diff
@@ -37,22 +37,35 @@ Face capture uses the webcam by default for capturing live photos of users. File
...
 
 For security reasons, instead of using the API token directly in you client-side code, you will need to generate and include a short-lived JSON Web Token (JWT)\* everytime you initialise the SDK. The easiest way to generate a JWT is via the [JWT endpoint](https://documentation.onfido.com/#json-web-tokens) in the Onfido API:
 
 ```shell
-$ curl https://api.onfido.com/v2/jwt?referrer=YOUR_WEBPAGE_URL \
+$ curl https://api.onfido.com/v2/jwt?referrer=YOUR_WEBPAGE_URL&applicant_id=YOUR_APPLICANT_ID \
```
hjheath: This is not the correct request to get a token - see the PR on the API docs

README.md
```diff
@@ -138,28 +151,27 @@ Congratulations! You have successfully started the flow. Carry on reading the ne
 - Handle callbacks
 - Remove the SDK from the page
 - Customise the SDK
-- Upload files to the Onfido API
 - Create checks
 
 ## Handling callbacks
 
-A number of callback functions are fired at various points of the flow. The most important function is `onComplete`. Inside this callback, you would typically send the captured images to your backend server, where you would then [upload files](#uploading-files-to-onfido-api) and [create checks](#creating-checks) using the [Onfido API](https://documentation.onfido.com/).
+A number of callback functions are fired at various points of the flow. The most important function is `onComplete`. Inside this callback, you would typically [create checks](#creating-checks) using the [Onfido API](https://documentation.onfido.com/) on your backend server.
```
hjheath: We should make it more clear you should not actually create a check from the client. e.g. "Inside this callback you would typically trigger your backend to create a check using the Onfido API."

README.md
```diff
@@ -138,28 +151,27 @@ Congratulations! You have successfully started the flow. Carry on reading the ne
...
 
 - **`onFaceCapture {Function} optional`**
 
-  Callback that fires when the face has been successfully captured and confirmed by the user. It returns an object that contains the face capture.
+  Callback that fires when the face has been successfully captured, confirmed by the user and uploaded to the Onfido API. It returns an object that contains the face capture.
 
 - **`onComplete {Function} optional`**
 
-  Callback that fires when both the document and face have successfully been captured. It returns an object that contains both captures. This event data should sent to your backend where the full [API requests](#uploading-files-to-onfido-api) will be made.
+  Callback that fires when both the document and face have successfully been captured and uploaded. It returns an object that contains both captures. This event data should sent to your backend where the full [API requests](#uploading-files-to-onfido-api) will be made.
```
hjheath: Remove references to uploading files.

README.md
```diff
@@ -170,51 +182,44 @@ A number of callback functions are fired at various points of the flow. The most
     containerId: 'onfido-mount',
     // here we send the data in the complete callback
     onComplete: function() {
-      var captures = Onfido.getCaptures()
-      sendToServer(captures)
+      startCheck(YOUR_APPLICANT_ID)
```
hjheath: We should not create checks from the browser!

src/components/Capture/capture.js
```diff
@@ -97,11 +98,15 @@ class Capture extends Component {
       return
     }
     payload = {...payload, documentType}
-    if (this.props.side === 'back' && !this.props.useWebcam) {
-      payload = {...payload, valid: true}
+    if (this.props.useWebcam) {
+      postToServer(this.createJSONPayload(payload), serverUrl, token,
+        (response) => this.onServerResponse(response),
```
rfreitas: because `onServerResponse` is an arrow method and therefore bound to this, you can just pass it directly as a parameter.

rfreitas: Same applies for `onServerError`

src/components/Confirm/index.js
```diff
@@ -81,18 +82,25 @@ const Previews = ({capture, retakeAction, confirmAction} ) =>
...
   return <Previews
     capture={capture}
     retakeAction={() => deleteCaptures({method, side})}
     confirmAction={() => {
-      confirmCapture({method, id: capture.id})
-      confirmEvent(method, side)
-      nextStep()
+      postToOnfido(capture, method, token, allowApiAdvancedValidation,
+        () => completeCapture(method, side),
+        (error) => onApiError(error)
```
rfreitas: `onApiError` can be passed directly, since it's bound to `this`

src/components/utils/errors.js
```diff
@@ -0,0 +1,8 @@
+export const errors = {
```
rfreitas: this is UI code specific to the sdk, it's not really an utility. Utilities then to be general purpose tools, not business specific.

rfreitas: you might want to create a module called strings, but I would keep it outside of any `utils` naming.

src/components/utils/onfidoApi.js
```diff
@@ -0,0 +1,58 @@
...
+  sendFile({blob, type: documentType, side, advanced_validation: allowApiAdvancedValidation}, 'documents', token, onSuccess, (request) => handleApiError(request, onError))
+}
+
+const objectToFormData = (object) => {
+  const formData = new FormData()
+  forEach(object, (value, key) => formData.append(key, value))
+  return formData
+}
+
+const sendFile = ({blob, ...extraOptions}, path, token, onSuccess, onError) => {
```
rfreitas: this is actually send File to onfido.

src/components/Uploader/index.js
```diff
@@ -17,35 +18,11 @@ const UploadProcessing = () =>
...
-const ServerError = ({message}) =>
-  <UploadError>{message}</UploadError>
-
-InvalidFileSize.defaultProps = {
-  message: 'The file size limit of 10 MB has been exceeded. Please try again.'
-}
-
-ServerError.defaultProps = {
-  message: 'There was an error connecting to the server. Please wait and try again later.'
+const showError = (uploading, error) => {
```
rfreitas: this should be a component, no reason for it to be a function. If it's component it's much more idiomatic

src/components/Capture/capture.js
rfreitas: the `onServerResponse` response inside is looking at the `valid` property. is this something the api is returning? Also, we should store the document/face id the api gives us and return that to the user instead.

src/components/utils/http.js
```diff
@@ -2,35 +2,37 @@ import { events } from '../../core'
...
-    initialValue)
-
 const handleError = ({status, response}, callback) => {
   console.error(status, response)
   Tracker.sendError(`${status} - ${response}`)
   callback()
 }
 
-export const postToServer = (payload, serverUrl, token, onSuccess, onError) => {
+export const postToServer = (payload, serverUrl, token, onSuccess, errorCallback) => {
```
rfreitas: can you put this in a sdk backend specific module? together with `SDK_SERVER_URL`

src/components/utils/onfidoApi.js
rfreitas: or make it consistent with the rest and remove onfido from the other functions.

hjheath:
The code looks good. I have some general questions:

- I think we should console log the API error message - it's easier to debug.
- Have you considered what needs to be done with surge to allow proper CI testing?

src/components/Capture/capture.js
```diff
@@ -26,7 +26,8 @@ class Capture extends Component {
     this.state = {
       uploadFallback: false,
       error: false,
-      hasWebcam: hasWebcamStartupValue
+      hasWebcam: hasWebcamStartupValue,
+      allowApiAdvancedValidation: true
```
hjheath: this name could be `advancedValidation` - shorter and consistent with the API

src/components/Capture/capture.js
stefaniacardenas: No, the `valid` property is only returned by the SDK backend and `onServerResponse` is only called when we receive a response from the server (SDK backend).
I don't return the document/face id because in order to run a document or facial similarity check the client doesn't need the doc/face id, and I agreed about this with See Wah during feature design. At the same time I agree with you on the fact that we are already including the id returned by the SDK backend in the documentCapture object and for consistency we could just replace it with the document id returned by the api. In that case, is it worth checking with Richard about potential security risks in returning the document id on the client? Anyways, I think this should be handled as part of another story because this isn't a blocker and I want to prioritise finishing the error feature for webcam screen during this sprint.

src/components/Capture/capture.js
```diff
@@ -97,11 +98,15 @@ class Capture extends Component {
...
+    if (this.props.useWebcam) {
+      postToServer(this.createJSONPayload(payload), serverUrl, token,
+        (response) => this.onServerResponse(response),
+        (response) => this.onServerError(response)
+      )
+      this.setState({allowApiAdvancedValidation: false})
     }
     else {
-      postToServer(this.createJSONPayload(payload), serverUrl, token, (response) => this.onServerResponse(response), (response) => this.onServerError(response))
+      payload = {...payload, valid: true}
```
hjheath: This feels redundant. Why do we need the concept of a capture being valid anymore given we track whether or not to set advanced_validation?

src/components/utils/http.js
stefaniacardenas: Yes, I was thinking about doing that but forgot to refactor it. Will do

webpack.config.babel.js
```diff
@@ -35,7 +35,10 @@ const basePlugins = [
   new webpack.NoErrorsPlugin(),
   new webpack.optimize.DedupePlugin(),
   new webpack.DefinePlugin({
-    'process.env.NODE_ENV': JSON.stringify(ENV)
+    'process.env': {
+      'NODE_ENV': JSON.stringify(ENV),
+      'ONFIDO_URL': JSON.stringify(process.env.ONFIDO_URL || 'https://apidev.onfido.com')
```
hjheath: This is not consistent with how we set the SDK URL.

Should it default to the staging environment?

src/components/utils/errors.js
stefaniacardenas: Yes, good point, it might come in handy when will work on the internationalisation.

src/components/Capture/capture.js
stefaniacardenas: Because if the capture isn't `valid` we cannot move to the confirm screen where we send it to the Onfido API to perform the "real validation". I guess you have a point, before we were only allowing the user to go to the confirm screen if the capture was valid but now the concept has changed so maybe we should rename that key.

hjheath: Can we remove it altogether?

stefaniacardenas: I have to look into it because this is using the sdk core selectors and I'm not sure how that works. 😬

webpack.config.babel.js
stefaniacardenas: Good point! Also because there is no way to store this variable on the backend as there is no backend... We pass the SDK url as a prop when we inject the SDK to then use it in the Capture component, I don't see the need to do this though and probably we should just return the right url based on `process.env.NODE_ENV` for both the SDK backend and API, instead of passing the prop from the ancestor to the children. What do you think?

hjheath: I think the reason we passed the URL as a prop was because that was the only way to propagate the value to core. Therefore we could now use Devise for both.

I think basing it on NODE_ENV is dangerous in case a customer is building the SDK in a development environment. Perhaps we could have an ONFIDO_ENV?

Unrelated I think you should namespace the Onfido API  interactions and the Onfido SDK server so we have consistent names for each throughout.

src/components/utils/onfidoApi.js
```diff
@@ -0,0 +1,58 @@
...
+  const bodyOptions = {
```
hjheath: we should set the sdk_source and sdk_version

hjheath:
There's a problem with the capture callbacks - we fire the callbacks when the user presses confirm, whereas the store is only updated with the image once the response comes back from the API (i.e. after the confirm). Therefore the callback returns `null` for the capture data.

src/components/Capture/capture.js
stefaniacardenas: The concept of valid captures is not really applicable anymore, captures should probably be marked as confirmable when they are suitable for confirmation from the user and ready to be sent to API. This requires quite a big refactoring on core therefore I won't be able to handle it as part of this story.

src/components/utils/onfidoApi.js
stefaniacardenas: Made it consistent by removing reference to Onfido in the other functions apart from the function that gets exported

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

webpack.config.babel.js
```diff
@@ -56,7 +65,7 @@ if (ENV === 'production') {
 
 const configDist = {
   context: `${__dirname}/src`,
-  entry: './index.js',
+  entry: ["babel-polyfill", './index.js'],
```
rfreitas: where is the babel-polyfill being used?

rfreitas: where is the babel-polyfill being used? this might have something to do with the 26% increase in size of the sdk.

stefaniacardenas: I've used `babel-polyfill` because it allows me to use some functionalities that are  normally not available on IE11. It would fix errors like `'Symbol' is undefined` and I would also be able to use methods like `includes` for arrays etc. 26% size increase is a lot, maybe I can look for a different solution

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

README.md
```diff
@@ -37,22 +37,35 @@ Face capture uses the webcam by default for capturing live photos of users. File
...
+    -H 'Authorization: Token token=YOUR_API_TOKEN' \
+    -d 'first_name=John' \
+    -d 'last_name=Smith'
+```
+
+You will receive a response containing the applicant id which will be used to create a JSON Web Token.
+
+### 3. Generating an SDK token
+
+For security reasons, instead of using the API token directly in you client-side code, you will need to generate and include a short-lived JSON Web Token ([JWT](https://jwt.io/)) every time you initialise the SDK. The easiest way to generate an SDK Token is via the [SDK Token endpoint](https://documentation.onfido.com/#sdk-tokens) in the Onfido API:
```
hjheath: It's not the easiest way - it's the only way!

README.md
```diff
@@ -37,22 +37,35 @@ Face capture uses the webcam by default for capturing live photos of users. File
...
+
+```shell
+$ curl https://api.onfido.com/v2/sdk_token?referrer=YOUR_WEBPAGE_URL&applicant_id=YOUR_APPLICANT_ID \
```
hjheath: This is not how you generate the tokens. See:  https://bitbucket.org/onfido/api-docs/pull-requests/112/update-sdk-token-section/diff

README.md
```diff
@@ -138,30 +151,29 @@ Congratulations! You have successfully started the flow. Carry on reading the ne
...
 
 - **`onFaceCapture {Function} optional`**
 
-  Callback that fires when the face has been successfully captured and confirmed by the user. It returns an object that contains the face capture.
+  Callback that fires when the face has been successfully captured, confirmed by the user and uploaded to the Onfido API. It returns an object that contains the face capture.
 
 - **`onComplete {Function} optional`**
 
-  Callback that fires when both the document and face have successfully been captured. It returns an object that contains both captures. This event data should sent to your backend where the full [API requests](#uploading-files-to-onfido-api) will be made.
+  Callback that fires when both the document and face have successfully been captured and uploaded. It returns an object that contains both captures. This event data should sent to your backend where the full [API requests](https://documentation.onfido.com/#create-check) will be made.
```
hjheath: This paragraph needs updating. They do not need to send any data to the backend - they just need to trigger the backend to create a check.

Also `both captures` should read `all captures` because there could be multiple documents

README.md
```diff
@@ -170,51 +182,47 @@ A number of callback functions are fired at various points of the flow. The most
     containerId: 'onfido-mount',
     // here we send the data in the complete callback
     onComplete: function() {
-      var captures = Onfido.getCaptures()
-      sendToServer(captures)
+      //communicate with your backend and initiate the check
```
hjheath: This still implies you could create a check from the client. What about `Tell your backend service that it can create the check`

README.md
```diff
@@ -170,51 +182,47 @@ A number of callback functions are fired at various points of the flow. The most
...
-      formData.append(key, value)
-      return formData;
-    }, new FormData())
+  ```
+  The `getCaptures` function contains the document and face files captured during the flow. Based on the applicant id, you can then create a check for the user via your backend. Here is an example of how to initiate a check from your backend.
 
-  const postData = (data, endpoint, callback) => {
-    const request = new XMLHttpRequest()
-    request.open('POST', endpoint, true)
+  ```js
```
hjheath: I don't think this example is necessary. The fact it is in JS also is confusing as it suggests this could be run on the client. I think we should just link to https://documentation.onfido.com/#create-check here

hjheath: In fact we already have that below so I would remove this section entirely.

README.md
```diff
@@ -359,7 +334,7 @@ $ curl https://api.onfido.com/v2/applicants/YOUR_APPLICANT_ID/checks \
     -d 'reports[][name]=facial_similarity'
```
hjheath: This example should include a document check

README.md
```diff
@@ -359,7 +334,7 @@ $ curl https://api.onfido.com/v2/applicants/YOUR_APPLICANT_ID/checks \
...
 ```
 
-Note: you can also submit the POST request in JSON format.
+Note: you can also submit the POST request in JSON format, see [example above](#handling-callbacks).
```
hjheath: I would remove this line

src/components/strings/errors.js
```diff
@@ -0,0 +1,8 @@
+export const errors = {
```
hjheath: Nice!

rfreitas:
Travis automatic deployment: https://staging-172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

README.md
stefaniacardenas: Oh yeah I forgot `-X POST` before the url, anyways for consistency I will use the same style we use on the API documentation

hjheath: No that's not the way to fix this. You need to pass the parameters in the body rather than the query string.

```
$ curl https://api.onfido.com/v2/sdk_token
  -H 'Authorization: Token token=<your_api_token>'
  -F 'applicant_id=<applicant_id>'
  -F 'referrer=<referrer_pattern>'```

Comment (stefaniacardenas) [inline README.md]: I know, as I said I will use the API style instead of the query string. However it does work even if I use query string instead of form data. Isn't query string just another way of specifying the body parameters? Or is there something I am missing?
The syntax would be
```shell
curl -X POST 'https://api.onfido.com/v2/sdk_token?referrer=<referrer_pattern>&applicant_id=<applicant_id>' -H 'Authorization: Token token=<your_api_token>'
```
Again, I will use the form data as in the API documentation, I'm just trying to understand what's wrong with using query string.

stefaniacardenas: https://stackoverflow.com/questions/10406588/api-requires-post-arguments-in-a-query-string

README.md
stefaniacardenas: It does

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-172-pr-onfido-sdk-ui-onfido.surge.sh

README.md
hjheath: The query string and body are two different parts of an HTTP request. They can both be used to send parameters and Rails parses them both when you access `params` in the controller. There are a few reasons why I think we should use the body here:

1) You have to do URL encoding in the query string: e.g. referrer `https%3A%2F%2F*.onfido.com%2Fdocumentation%2F*`. Sending it in the body is much more readable.
2) We should be consistent with the rest of the API. Whilst this endpoint is not strictly RESTful, our API as a whole is. Developers expect the URL + query string to identify a resource and the body to contain the data to upload to the server. In this case the referrer and applicant ID are uploaded data rather than resources identifiers.
3) It makes it obvious it's a POST request rather than a GET request. GET requests must have an empty body. Sending a POST request with an empty body is a slightly weird use case - notice `curl` does a GET by default with no body and POST by default with a body.
4) Security - the whole request URL will get logged in loads of places. Therefore you need to be careful what you put in it. Typically you have more control over whether the body parameters get logged.

README.md
```diff
@@ -138,30 +153,29 @@ Congratulations! You have successfully started the flow. Carry on reading the ne
...
-  Callback that fires when the face has been successfully captured and confirmed by the user. It returns an object that contains the face capture.
+  Callback that fires when the face has been successfully captured, confirmed by the user and uploaded to the Onfido API. It returns an object that contains the face capture.
 
 - **`onComplete {Function} optional`**
 
-  Callback that fires when both the document and face have successfully been captured. It returns an object that contains both captures. This event data should sent to your backend where the full [API requests](#uploading-files-to-onfido-api) will be made.
+  Callback that fires when both the document and face have successfully been captured and uploaded. It returns an object that contains all captures. At this point you can trigger your backend to create a check by making a request to the [create check endpoint](https://documentation.onfido.com/#create-check)(https://documentation.onfido.com/#create-check).
 
-  Here is an `onComplete` callback example, making use of the `getCaptures` convenience function:
+  Here is an `onComplete` callback example:
```
hjheath: I'm not sure if we should keep this example given that it's empty

rfreitas: not sure I follow, what is empty?

rfreitas:
@dawiddylowicz can you do a pre merge acceptance?

README.md
stefaniacardenas: Cool! Thanks, that's a good explanation. I've updated it. I had experience with APIs that use query string in the past that's why I was using the query string.
Although, I don't think that the parsing of the query string into body is a Rails specific thing but rather it's how it gets included by any service using tcp. Or at least that was my understanding from this article https://en.wikipedia.org/wiki/Query_string
`The form content is only encoded in the URL's query string when the form submission method is GET. The same encoding is used by default when the submission method is POST, but the result is submitted as the HTTP request body rather than being included in a modified URL.`

README.md
hjheath: the callback example function

rfreitas: I think it helps, you can see in the example where you should write your own logic

README.md
```diff
@@ -138,30 +153,29 @@ Congratulations! You have successfully started the flow. Carry on reading the ne
...
 - Customise the SDK
-- Upload files to the Onfido API
 - Create checks
 
 ## Handling callbacks
 
-A number of callback functions are fired at various points of the flow. The most important function is `onComplete`. Inside this callback, you would typically send the captured images to your backend server, where you would then [upload files](#uploading-files-to-onfido-api) and [create checks](#creating-checks) using the [Onfido API](https://documentation.onfido.com/).
+A number of callback functions are fired at various points of the flow. The most important function is `onComplete`. Inside this callback you would typically trigger your backend to create a check using the [Onfido API](https://documentation.onfido.com/).
 
 - **`onReady {Function} optional`**
```
rfreitas: we should remove this callback though, since it's no longer working. see https://github.com/onfido/onfido-sdk-ui/issues/153

README.md
```diff
@@ -170,51 +184,13 @@ A number of callback functions are fired at various points of the flow. The most
     containerId: 'onfido-mount',
```
rfreitas: @hjheath we probably don't need the ID though in this example. Also might be helpful to write the arguments which are passed to it.

hjheath: I agree, let's do it outside of this PR though

dawiddylowicz:
@rfreitas yup, already synced with @stefaniacardenas to check that one. Tomorrow morning is the time, probably?

stefaniacardenas:
Yes, I've pushed the last commit so you can start whenever you want. Thanks @dylowicz

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-172-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://172-pr-onfido-sdk-ui-onfido.surge.sh

dawiddylowicz:
PR ready to merge:

* automated tests passing

* regression tests passing

* exploratory tests passed

* bugs fixed

README.md
```diff
@@ -37,22 +37,37 @@ Face capture uses the webcam by default for capturing live photos of users. File
...
+Make a note of the `token` value in the response, as you will need it later on when initialising the SDK.
 
-\* Tokens expire 24 hours after creation. See [here](https://jwt.io/) for details on how JWTs work.
+\* Tokens expire 90 minutes after creation.
 
-### 3. Including/Importing the library
+### 4. Including/Importing the library
 
-#### 3.1 HTML Script Tag Include
+#### 4.1 HTML Script Tag Include
```
dawiddylowicz: Correct the points' numbers

* other`3.2 NPM style import` subpoint from 4. is still not updated

* also next point is `4. Adding basic HTML markup` when should be 5

PR #177, opened 2017-07-20
jsFiddle example: using master/dist; onReady removed

rfreitas:
## Problem

The jsFiddle example was using the onReady callback.
Also it was using the surge dist, which might be different from the dist in github, which is the one developers will actually consume.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ And the CSS styles:
 
 #### Example app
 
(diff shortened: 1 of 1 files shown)


rfreitas:
Travis automatic deployment: https://177-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://177-pr-onfido-sdk-ui-onfido.surge.sh

PR #200, opened 2017-08-25
Add cucumber tests to repository

dawiddylowicz:
* Add Cucumber tests to repository, using a private `monster` gem
* Setup Travis CI to run Cucumber tests on `localhost` after each build on headful Chrome browser
* Add script to install latest beta Google Chrome and chromedriver on Travis CI


rfreitas:
Travis automatic deployment: https://staging-200-pr-onfido-sdk-ui-onfido.surge.sh

test.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
dawiddylowicz: For now this script will work only on CI, not yet ready to be run locally for testing. Separately, I'll create a guide how to setup and run tests locally, which is quite simple to be honest.

test.sh
```diff
@@ -0,0 +1,28 @@
...
+sed '/webpack: Compiled successfully.$/q' <&3 ; cat <&3 &
+
+# go to test directory
+cd $TESTS_PATH
+
+# install gem dependencies using monster_rsa private key to fetch the monster gem
+GIT_SSH_COMMAND="ssh -i ~/.ssh/monster_rsa" bundle install
+
+# run cucumber tests against deployed domain
+SDK_SURGE_URL="https://localhost:8080/?async=false"
```
rfreitas: it's no longer the surge link, right?

.travis.yml
```diff
@@ -9,10 +11,19 @@ env:
 cache:
   directories:
     - "node_modules"
+before_install:
+  - chmod ugo+x ./test/install_chromedriver_travis.sh
```
rfreitas: I think this can only wrapped into one script, maybe called `test_environment_setup.sh`?

test.sh
dawiddylowicz: right :) will change

.travis.yml
dawiddylowicz: good idea

dawiddylowicz:
`test/monster_rsa.enc` is an encrypted private ssh key with only read access to private monster gem repository. This is required in order to allow Travis to fetch the dependency. The solution has been reviewed upfront by security team

dawiddylowicz:
Tests are occasionally failing due to unstable environment on Travis. I'm still investigating the cause but my bets are:
* browser boot-up delays - tests are trying to be executed before the browser loads
* tests not using the latest Chrome (even though the script explicitly installs and links latest beta version on Travis)
* local webpack server eventually runs out of open connections

The quickest solutions to solve this would be:
* retry failed tests
then we could
* find out how to force `selenium-webdriver` using the latest, installed version of chrome (I've already spent some hours on trying to get this done)
* investigate if webpack server runs out of connections and how to prevent

Worth to add that tests are really stable when ran locally.

rfreitas:
Travis automatic deployment: https://200-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://200-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://200-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-200-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://200-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-200-pr-onfido-sdk-ui-onfido.surge.sh

dawiddylowicz:
I've addressed mentioned issues and improved tests stability. So that now:
* test waits 30 seconds for page to load
* test waits 30 seconds for elements assertions (i.e. to be visible)
* tests are now using the latest stable version of Chrome
* tests will retry twice after a failed scenario, before eventually failing the whole suite

Build is green, please continue on reviewing!

.travis.yml
```diff
@@ -9,10 +11,16 @@ env:
 cache:
   directories:
     - "node_modules"
+before_install:
+  - openssl aes-256-cbc -K $encrypted_1688a3f2e0a8_key -iv $encrypted_1688a3f2e0a8_iv -in test/monster_rsa.enc -out ~/.ssh/monster_rsa -d
```
rfreitas: Minor: any reason why this cannot be inside of `setup_test_environment`?

dawiddylowicz: Actually I tried this and it simply didn't work for some reason. Plus that line was automatically added to this step by Travis CLI, according to this: https://docs.travis-ci.com/user/encrypting-files/

test.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -e
+
+# set path for test directory
+TESTS_PATH=./test
```
hjheath: Di you consider using travis for environment variables? (Although it may make our travis file too verbose)

test.sh
```diff
@@ -0,0 +1,28 @@
...
+# ref: https://stackoverflow.com/a/21002153
+echo "Running local server..."
+exec 3< <(npm run dev)
+sed '/webpack: Compiled successfully.$/q' <&3 ; cat <&3 &
+
+# go to test directory
+cd $TESTS_PATH
+
+# install gem dependencies using monster_rsa private key to fetch the monster gem
+GIT_SSH_COMMAND="ssh -i ~/.ssh/monster_rsa" bundle install
```
hjheath: Out of interest what version of ruby are we using in travis?

test/scripts/install_chromedriver_travis.sh
```diff
@@ -0,0 +1,30 @@
...
+echo "Uninstalling current Chromium from `which chromium-browser`..."
+sudo apt-get purge chromium-browser
+rm ~/.config/chromium/ -rf
+echo "Uninstalling current Google Chrome from `which google-chrome`..."
+sudo apt-get purge google-chrome-stable
+rm ~/.config/google-chrome/ -rf
+
+echo "Installing $CHROME_VERSION..."
+sudo wget --no-verbose -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -
+echo "deb http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google-chrome.list
```
hjheath: This log doesn't make sense to me - can you explain?

test/scripts/install_chromedriver_travis.sh
```diff
@@ -0,0 +1,30 @@
...
+sudo apt-get update -qqy
+sudo apt-get -qqy install ${CHROME_VERSION}
+sudo rm /etc/apt/sources.list.d/google-chrome.list
+sudo rm -rf /var/lib/apt/lists/* /var/cache/apt/*
+sudo ln -fs /opt/google/chrome-beta/google-chrome /usr/bin/google-chrome
+sudo ln -fs /opt/google/chrome-beta/google-chrome /usr/local/bin/google-chrome
```
hjheath: Can you explain these 6 lines?

test/scripts/install_chromedriver_travis.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
```
hjheath: Why do we need to install a different version of chrome?

dawiddylowicz: Default Chrome on Travis is of version 59 which is more likely to cause problems with chromedriver during tests. That's why I'm first making sure there are no chrome browsers on Travis, and then installing the latest stable one - this will be especially important when we move to running tests on Chrome in headless mode

hjheath:
Looks really good. A few questions which are probably out of scope of this PR:

- Do you know why the dist has increased in size?
- Is the 30 seconds timeout and double retry excessive? Can we improve the stability of the tests and so reduce the time they take to run.
- Can we improve the travis log - on this page we only get pass/fail and on the travis details we get very verbose logs. It would be great to get a pipeline view of what failed (build vs tests vs deploy)

test/scripts/install_chromedriver_travis.sh
dawiddylowicz: This was initially taken from an implementation of Dockerfile in our cucumber repository which was driven for purposes of running tests on headless Chrome that required latest beta version at the time (https://bitbucket.org/onfido/cucumber/src/2adc5cb302a0db505bc8d165e81005e4c5e25d95/Dockerfile?at=master&fileviewer=file-view-default). Since we're currently not using headless Chrome due to an issue on Travis + the browser is now stable, we could delete some of these lines - I'll experiment a little to see if it still works

test.sh
dawiddylowicz: I did not in this case. This variable is used only within this script so I found it more clear to keep it there

test.sh
dawiddylowicz: We're using a default one that's installed on Travis. Just checked and apparently it's quite an old version we should consider updating:
`ruby 1.9.3p484 (2013-11-22 revision 43786) [x86_64-linux]`

dawiddylowicz:
@hjheath Thanks for review. I pushed changes according to your comments.
* No idea about the dist, I'm not entirely sure how it's calculated and what's the baseline for that calculation, but may it be because the branch is not synced completely with current development state?
* It might be excessive but for now I would just call it 'safer'. If tests are healthy, this does not impact execution time in any way. It may only prolong execution time if scenarios are failing - we could fix that by failing fast, for example.
* We can improve Travis log for sure. I couldn't find a simple way to report tests separately than the job log so I just left it this way. Also the pipeline thing sounds interesting - something we can definitely think about in near future.

rfreitas:
Travis automatic deployment: https://staging-200-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://200-pr-onfido-sdk-ui-onfido.surge.sh

test/features/sdk_flow.feature
```diff
@@ -0,0 +1,203 @@
...
+    Then error_message () should contain "Unsupported file type"
+    And error_instruction () should contain "Try using a .jpg or .png file"
+
+  Scenario: I should not be able to upload an image containing multiple faces
+    Given I navigate to the SDK
+    When I click on verify_identity (SDK)
+    And I click on passport ()
+    Then page_title () should contain "Upload the front of your document"
+    When I upload passport on file_upload ()
+    Then I should see uploaded_image ()
```
stefaniacardenas: Is it possible to group steps to avoid repetitions? For example, could you group the steps from
`When I click on verify_identity (SDK)` to `Then I should see uploaded_image ()` in one step called "When I successfully upload a passport" ?

dawiddylowicz:
Added following changes:
*`--retry` now works properly
* Travis now runs local webpack server against `production`
* tests are now fired only on `NODE_ENV=production` builds
* tests are now fired only on PRs, not on push

rfreitas:
Travis automatic deployment: https://staging-200-pr-onfido-sdk-ui-onfido.surge.sh

test/features/sdk_flow.feature
dawiddylowicz: Absolutely! I was thinking of that too as there are many repetitive pieces of code. Though I would rather do it as part of other PR

rfreitas:
Travis automatic deployment: https://200-pr-onfido-sdk-ui-onfido.surge.sh

PR #213, opened 2017-09-22
fix for surge disabling paths

rfreitas:
Surge was giving not found errors when we navigated to the mobile link.
This is probably due to surge blocking paths, which we were using to specify the room id.
The room is has been moved temporarily to the query parameters


rfreitas:
Travis automatic deployment: https://staging-213-pr-onfido-sdk-ui-onfido.surge.sh

PR #215, opened 2017-09-26
Feature/mobile linking ui cx 1176

stefaniacardenas:
In this PR, I implemented the component that allow the user to start the cross device flow and the screen where the mobile link can be copied. I refactored the Router component to allow switching between the default flow (called it "master") and the cross device one, in which the mobile link can be found as well as the screens displayed when the mobile is connected (in progress and complete)


rfreitas:
Travis automatic deployment: https://staging-215-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-215-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Router/StepComponentMap.js
```diff
@@ -31,6 +33,25 @@ const createDocumentComponents = (documentType) => {
   return frontDocumentFlow
 }
 
+
+const crossDeviceComponents = {
+  mobileLink: () => [MobileLink],
+  mobileConnection: () => [MobileFlowInProgress, MobileFlowComplete]
```
hjheath: This should be one component rather than two right? It's fine if you plan to fix as part of CX-1148

src/components/Router/index.js
```diff
@@ -1,22 +1,16 @@
...
 
-import { unboundActions, events } from '../../core'
-import {sendScreen} from '../../Tracker'
-import {wrapArray} from '../utils/array'
-import { createComponentList } from './StepComponentMap'
-
-const history = createHistory()
-
+import {componentsList} from './StepComponentMap'
+import { unboundActions } from '../../core'
```
hjheath: Is this import used?

src/components/Router/Flow.js
```diff
@@ -0,0 +1,91 @@
+import { h, Component } from 'preact'
+import createHistory from 'history/createBrowserHistory'
+
+import { events } from '../../core'
+import {sendScreen} from '../../Tracker'
+import {wrapArray} from '../utils/array'
+
+const history = createHistory()
+
+class Flow extends Component {
```
hjheath: I don't like this component name - I would prefer StepsRouter or LinearRouter.

I also think the names of the 'flows' i.e. 'master' and 'crossDevice' could be made clearer. crossDevice is OK, but I think master could be called 'upload' or something more descriptive

src/components/Router/Flow.js
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }
+
+  previousStep = () => {
+    const currentStep = this.state.step
+    this.setStepIndex(currentStep - 1)
+  }
+
+  changeFlow = () => {
+    const flow = this.props.nextFlow()
```
hjheath: Nice

src/components/Router/Flow.js
```diff
@@ -0,0 +1,91 @@
...
+import {wrapArray} from '../utils/array'
+
+const history = createHistory()
+
+class Flow extends Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      step: this.props.step,
+      componentsList: this.props.componentsList
```
rfreitas: doesn't seem like this is needed. this state property is always equal to props property right?

src/components/Router/index.js
```diff
@@ -57,7 +51,7 @@ class MobileRouter extends Component {
 
   render = (props) =>
       this.state.token ?
-        <StepsRouter {...props} {...this.state}/> : <p>LOADING</p>
+        <MasterFlow {...props} {...this.state}/> : <p>LOADING</p>
```
hjheath: This component no longer exists - we should render a Flow component here

src/components/Router/Flow.js
```diff
@@ -0,0 +1,91 @@
...
+    }
+    this.unlisten = history.listen(({state}) => {
+      this.props.onStepChange(state)
```
rfreitas: important to bear in mind that is the `onStepChance` props change, this is code is incomplete, since it doesn't take into account that change. Now taking into account props changing can lead no bugs down the line.

src/components/strings/uploadCopy.js
```diff
@@ -0,0 +1,19 @@
+export const desktopCopy = {
+  instructions: {
+    document: {
+      front: 'Upload front of document from your computer',
```
hjheath: This copy does not appear on the upload screen

src/components/crossDevice/SwitchDevice/style.css
```diff
@@ -0,0 +1,54 @@
+.container {
+  height: 96px;
```
hjheath: What did we conclude on pixels vs rem? I wasn't part of the debate, but you seem to be using them interchangeably

hjheath:
Could you squash the commit history before merging? In particular the commits around fixing the history are a bit verbose

src/components/Router/Flow.js
rfreitas: so `changeFlow` triggers a toggle between flows, correct? If so, it makes it a bit less explicit when reading the code what happens next. Did you consider `changeToFlow(flowName)` instead?

src/components/Router/StepComponentMap.js
```diff
@@ -31,6 +33,25 @@ const createDocumentComponents = (documentType) => {
...
+}
+
+const mobileSteps = [{'type': 'mobileLink'}, {'type': 'mobileConnection'}]
```
rfreitas: hmm, hadn't fully considered that the tracking requires a developer step.

rfreitas:
some heavy lifting going on : > first skim impression: looks simple (as much as we can make it for now at least)!

rfreitas:
Travis automatic deployment: https://staging-215-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Uploader/style.css
```diff
@@ -43,9 +42,11 @@
   cursor: pointer;
   @media (--large) {
     min-height: 20rem;
+    margin: 1rem;
+    min-height: 20rem;
+    height: 264px;
```
hjheath: Could you check that the new UI matches the design? In this case the dropzone is too large because of the `min-height` (which is also repeated)

src/components/crossDevice/MobileLink/index.js
```diff
@@ -0,0 +1,56 @@
...
+    return (
+      <div>
+        <div className={style.container}>
+          <h1 className={`${theme.title} ${style.title}`}>Continue verification on your mobile</h1>
+          <div>Copy and send the below link to your mobile</div>
+
+          <div className={style.linkSection}>
+            <div className={style.linkTitle}>Secure link</div>
+            <div className={classNames(style.actionContainer, {[style.copySuccess]: this.state.copySuccess})}>
+              <textarea ref={(textarea) => this.textArea = textarea} value={mobileUrl} />
```
hjheath: The link is wrapping out of sight when it is too long. In production the link should be shorter, but we should make sure it never disappears from view.

src/components/Router/Flow.js
rfreitas: or `zeus` since he is the master of gods

rfreitas: joking...

src/components/Router/StepComponentMap.js
stefaniacardenas: Yes, it's outside the scope of this ticket, I've only added the components to test the navigation

rfreitas:
Travis automatic deployment: https://staging-215-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Router/Flow.js
stefaniacardenas: Yes, I agree. `StepsRouter` makes sense instead of `Flow` and I will replace 'master' with 'captureSteps' and 'crossDevice' with 'crossDeviceSteps'

src/components/Router/Flow.js
rfreitas: I was wrong about this!

src/components/Router/index.js
stefaniacardenas: Yes, in `mapDispatchToProps()`

src/components/Uploader/style.css
stefaniacardenas: Good spot, I have to double check all the new changes that Daniel have made to the design after I have already implementing the original design. Will update it

src/components/crossDevice/MobileLink/index.js
stefaniacardenas: Yes, you're right. I was hoping for a shorter link that's why I didn't make the textarea expandable but I will make sure that link is always fully visible

src/components/Uploader/style.css
stefaniacardenas: There is a ticket for that in Jira https://jira.onfido.co.uk/browse/CX-1217
We haven't started the debate yet. We can use both rem and px at the same time, it's not a problem, as long as we do it properly so that the UI looks good when a user changes the front size in their browser settings. I wonder what's the percentage of users that change the font size in their browser settings, and if it's big enough maybe we should tackle this issue as part of the accessibility epic, whenever we will work on that.

src/components/crossDevice/SwitchDevice/style.css
stefaniacardenas: There is a ticket for that in Jira https://jira.onfido.co.uk/browse/CX-1217
We haven't started the debate yet. We can use both rem and px at the same time, it's not a problem, as long as we do it properly so that the UI looks good when a user changes the front size in their browser settings. I wonder what's the percentage of users that change the font size in their browser settings, and if it's big enough maybe we should tackle this issue as part of the accessibility epic, whenever we will work on that.

src/components/strings/uploadCopy.js
stefaniacardenas: Oh yes, it was due to the merging

src/components/Uploader/index.js
```diff
@@ -4,22 +4,29 @@ import theme from '../Theme/style.css'
...
 const UploadError = ({error}) =>
   error && <div className={`${style.text} ${style.error}`}>{`${error.message}. ${error.instruction}.`}</div>
 
-const MobileLink = ({mobileUrl}) =>
-  mobileUrl && <p className={style.mobileUrl}>Mobile: {mobileUrl}</p>
-
-const UploaderPure = ({onImageSelected, error, mobileUrl}) =>
+const UploaderPure = ({method, side, onImageSelected, error, ...{changeFlow}}) =>
   <div>
+    { isDesktop ? <SwitchDevice {...{changeFlow}} /> : '' }
```
rfreitas: Not sure if `isDesktop` is the most correct way of conditionally showing/not showing this.
Since it's not very reliable and reality is, we never want to show this if you are on cross device link.

rfreitas:
Travis automatic deployment: https://staging-215-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Document/index.js
```diff
@@ -19,11 +19,11 @@ DocumentTitle.defaultProps = {
   title: {
     front: {
       captureTitle: `Front of document`,
-      uploadTitle: `Upload the front of your document`
+      uploadTitle: `Upload front of document`
```
dawiddylowicz: Tests assertions need to be updated due to this change

src/components/Document/index.js
```diff
@@ -19,11 +19,11 @@ DocumentTitle.defaultProps = {
...
     },
     back: {
       captureTitle: `Back of document`,
-      uploadTitle: `Upload the back of your document`
+      uploadTitle: `Upload back of document`
```
dawiddylowicz: Same as above, tests assertions need to be updated

dawiddylowicz:
The arrow in "Need to use your mobile...?" box is misaligned on both Firefox and Safari

Firefox:
firefox

Safari:
safari

rfreitas:
Travis automatic deployment: https://staging-215-pr-onfido-sdk-ui-onfido.surge.sh

seewah:
Also, perhaps not explicitly defined, but I think the whole of the "Need to ..." mobile link box should be clickable, and not just the icon, or the arrow

rfreitas:
Travis automatic deployment: https://staging-215-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Router/StepsRouter.js
```diff
@@ -0,0 +1,95 @@
...
+import { events } from '../../core'
+import {sendScreen} from '../../Tracker'
+import {wrapArray} from '../utils/array'
+
+const history = createHistory()
+
+class StepsRouter extends Component {
+  constructor(props) {
+    super(props)
+    this.state = {
```
rfreitas: this state is not needed, both `step` and `componentsList` always equals the props

src/components/Router/index.js
```diff
@@ -82,105 +103,40 @@ class DesktopRouter extends Component {
...
-          trackScreen = {this.trackScreen}/>
-      </div>
+      <StepsRouter {...props}
+        componentsList={components}
+        flow={this.state.flow}
+        nextFlow={this.nextFlow}
+        step={this.state.step}
+        onStepChange={this.onStepChange}
+        roomId={this.state.roomId}
+        isMobileFlow={props.options.mobileFlow}
```
rfreitas: `isMobileFlow` is always `false` for `DesktopRouter` and `true` for `MobileRouter`. By reading those two components in isolation one can wonder if that might not be the case.
I would recommend setting `isMobileFlow` in the `Router`, so these two children (mobile and desktop) who shouldn't control that value just pass it down.

rfreitas:
Travis automatic deployment: https://staging-215-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-215-pr-onfido-sdk-ui-onfido.surge.sh

stefaniacardenas:
@seewah 
1) the whole of the "Need to ..." mobile link box should be clickable, and not just the icon, or the arrow - **Done**
2) "mobile users should not see the link" at all - is there anything we can do now to hide it for users seeing the flow on mobile/small screen devices, we can discuss - **In my first implementation it was working as I was using browser/device detection, yesterday I pushed a change and forgot to fix mobile. Ricardo is on it.**
3) the URL inside the copy link box should not break into 2 lines - *If we want to respect the original design, maximum I can do now is preventing the text from wrapping and reducing the font-size. Otherwise I need to expand the text area and button which is not trivial... depends how long the production link will be.*
4) secure link page breaks for small screen device, but maybe this is OK depending on what we want to do with point 2 - *I don't think we want to display the link on mobile devices that's why I didn't implement the small screen view.*
5) mobile link box right padding is lacking on FF - *Yes, @dawiddylowicz reported it yesterday. On it.*
6) safari layout problem as reported by @dawiddylowicz - *Yep*
7) surge link does not load on my VirtualBox IE at all :( - *I don't think this is due to my PR*
8) top padding inside the "how do i do this box" is a bit off compared with Invision - 😱 
9) in Invision the OPTION text is 14px, not 16px - *Two different font sizes in the same line?! I'll have a chat with Daniel when he is back* 😉

src/components/Router/index.js
rfreitas: working on this as part of https://github.com/onfido/onfido-sdk-ui/pull/217

rfreitas:
Travis automatic deployment: https://staging-215-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-215-pr-onfido-sdk-ui-onfido.surge.sh

stefaniacardenas:
Fixed all the above apart from:
8) top padding inside the "how do i do this box" is a bit off compared with Invision 
If I put a padding-top of 24px it won't look right unless I change the height of the container. I think it looks good enough the way it is tbh

seewah:
more feedback:

1. copy should say "Securely continue your verification on your mobile" and not "Safely"

rfreitas:
Travis automatic deployment: https://staging-215-pr-onfido-sdk-ui-onfido.surge.sh

stefaniacardenas:
This is how it would look it if I reduce the p tags margin and I put 24px margin top and 16px margin left + 0.35 opacity. Looks a bit transparent to me. What shall I do?
screen shot 2017-09-29 at 16 22 22

rfreitas:
Travis automatic deployment: https://staging-215-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-215-pr-onfido-sdk-ui-onfido.surge.sh

seewah:
incidentally, you should be able to put opacity on the border itself using a an alpha value no? but anyway I am happy with what we currently have :)

PR #227, opened 2017-10-06
Fix sending steps to cross device client

hjheath:
Bad bug, not sure when I introduced it!

### DIFF
--- a/src/index.ejs
+++ b/src/index.ejs
@@ -48,12 +48,6 @@
         if (options.async === "false") {
           getToken(createSDK)
         }
-        else if (options.token) {
-          createSDK(options.token)
-        }
-        else if (options.mobileFlow) {
-          window.onfidoOut = Onfido.init({mobileFlow: true})
-        }
         else {
           createSDK(null)
           getToken(injectToken)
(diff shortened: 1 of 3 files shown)


rfreitas:
Travis automatic deployment: https://227-pr-onfido-sdk-ui-onfido.surge.sh

src/index.ejs
```diff
@@ -48,12 +48,6 @@
         if (options.async === "false") {
           getToken(createSDK)
         }
-        else if (options.token) {
-          createSDK(options.token)
-        }
-        else if (options.mobileFlow) {
-          window.onfidoOut = Onfido.init({mobileFlow: true})
```
stefaniacardenas: So where do you set `mobileFlow` to `true` now?

hjheath: This is set in the hosted_web_sdk repo: https://bitbucket.org/onfido/hosted_web_sdk/src/4547923844801f855109d14892f0a73ac1efb334/production.html?at=master&fileviewer=file-view-default

This index.ejs is only used for development. The code I removed here is legacy from the proof of concept

stefaniacardenas: Oh yes, forgot about that.

PR #230, opened 2017-10-11
Pass woopra cookie between devices

hjheath:
This PR allows us to track users between devices on the cross device flow.

We use the woopra cookie as a unique identifier see: https://docs.woopra.com/v2.4/docs/profile-id-system#section-identifiers-in-woopra

When you open the mobile link we send the cookie from the desktop session and set it in your mobile session.

The JS docs advise using an email address as a unique identifer, but give we don't have one then using that field with a random ID would be a hack.

Setting the cookie is usually handled under the covers by the JS SDK, but it's OK to override that behaviour e.g. they do it in their Rails SDK https://www.woopra.com/docs/setup/ruby-on-rails-sdk/

### DIFF
--- a/src/components/Router/index.js
+++ b/src/components/Router/index.js
@@ -9,6 +9,7 @@ import StepsRouter from './StepsRouter'
 import Spinner from '../Spinner'
 import { unboundActions } from '../../core'
 import { isDesktop } from '../utils'
+import { getWoopraCookie, setWoopraCookie } from '../../Tracker'
 
 const history = createHistory()
 
@@ -39,20 +40,9 @@ class CrossDeviceMobileRouter extends Component {
 
   setConfig = (actions) => (data) => {
     const {token, steps, documentType, step, woopraCookie} = data
-    this.setWoopraCookie(woopraCookie)
+    setWoopraCookie(woopraCookie)
     this.setState({token, steps, step})
     actions.setDocumentType(documentType)
-  }
-
-  setWoopraCookie = (cookie) => {
-    const cookie_name = window.onfidojssdkwoopra.config('cookie_name')
-    const cookie_expire = window.onfidojssdkwoopra.config('cookie_expire')
-    const cookie_path = window.onfidojssdkwoopra.config('cookie_path')
-    const cookie_domain = window.onfidojssdkwoopra.config('cookie_domain')
-    window.onfidojssdkwoopra.docCookies.setItem(
-      cookie_name, cookie, cookie_expire, cookie_path, cookie_domain
-    )
-    window.onfidojssdkwoopra.cookie = cookie
   }
 
   onStepChange = ({step}) => {
@@ -85,7 +75,7 @@ class MainRouter extends Component {
   mobileConfig = () => {
     const {documentType, options} = this.props
     const {steps, token} = options
-    const woopraCookie = window.onfidojssdkwoopra.cookie
+    const woopraCookie = getWoopraCookie()
     return {steps, token, documentType, step: this.state.mobileInitialStep, woopraCookie}
   }
 
--- a/src/components/Router/index.js
+++ b/src/components/Router/index.js
@@ -38,9 +38,21 @@ class CrossDeviceMobileRouter extends Component {
   }
 
   setConfig = (actions) => (data) => {
-    const {token, steps, documentType, step} = data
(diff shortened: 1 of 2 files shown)


src/components/Router/index.js
```diff
@@ -41,11 +41,23 @@ class CrossDeviceMobileRouter extends Component {
...
+  setWoopraCookie = (cookie) => {
+    const cookie_name = window.onfidojssdkwoopra.config('cookie_name')
+    const cookie_expire = window.onfidojssdkwoopra.config('cookie_expire')
+    const cookie_path = window.onfidojssdkwoopra.config('cookie_path')
+    const cookie_domain = window.onfidojssdkwoopra.config('cookie_domain')
+    window.onfidojssdkwoopra.docCookies.setItem(
+      cookie_name, cookie, cookie_expire, cookie_path, cookie_domain
+    )
+    window.onfidojssdkwoopra.cookie = cookie
+  }
```
hjheath: cf. how the woopra JS SDK sets the cookie: https://github.com/onfido/js-client-tracker/blob/master/wpt.js#L885

rfreitas:
Travis automatic deployment: https://staging-230-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://230-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-230-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://230-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Router/index.js
```diff
@@ -39,11 +39,23 @@ class CrossDeviceMobileRouter extends Component {
...
 
   setConfig = (actions) => (data) => {
-    const {token, steps, documentType, step} = data
+    const {token, steps, documentType, step, woopraCookie} = data
+    this.setWoopraCookie(woopraCookie)
     this.setState({token, steps, step})
     actions.setDocumentType(documentType)
   }
 
+  setWoopraCookie = (cookie) => {
```
rfreitas: it's much better to do this inside of the Tracker module. It's specific to tracking and not the router. Also, I think you will be able to reference the onfido woopra object without having to access `window`, which should be avoided as much as possible.

hjheath: ok makes sense from a code organisation point of view. Why is accessing `window` bad?

rfreitas: because it's a global variable. Anyone can potentially rewrite that property to something else.

stephencookdev:
Are you sure that this is the best approach?

Messing about with woopra's internals is also a bit hacky, imo - and runs the risk of woopra changing their code and this breaking

IMO, setting the email to a random id is also a hack, but seems a more future-proof one

rfreitas:
Travis automatic deployment: https://staging-230-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://230-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Router/index.js
```diff
@@ -74,7 +76,8 @@ class MainRouter extends Component {
   mobileConfig = () => {
     const {documentType, options} = this.props
     const {steps, token} = options
-    return {steps, token, documentType, step: this.state.mobileInitialStep}
+    const woopraCookie = window.onfidojssdkwoopra.cookie
```
rfreitas: this reference to window can still be removed, by having a `getWoopraCookie` function.

hjheath:
@stephencookdev I decided against using the email as we would have to use a random ID (probably the cookie value) rather than an actual email address. Given I could set the cookie using public methods of their SDK it feels like less of a hack.

src/components/Router/index.js
hjheath: fixed

rfreitas:
Travis automatic deployment: https://230-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-230-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://230-pr-onfido-sdk-ui-onfido.surge.sh

PR #232, opened 2017-10-19
Feature/generic error screen cx 1232

stefaniacardenas:
In this PR we introduce a generic error to be displayed on the Cross Device Client if:
* something goes wrong on the master
* master isn't available
* client do not receive configuration
* token expires
* token expires mid-flow (ie. after you click on 'confirm' while on client)

I have also changed the CrossDeviceLink component to only generate one link


rfreitas:
Travis automatic deployment: https://staging-232-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Router/index.js
```diff
@@ -28,23 +30,37 @@ class CrossDeviceMobileRouter extends Component {
...
     this.requestConfig()
   }
 
+  componentDidMount() {
+    this.state.socket.on('disconnect', this.setError)
+  }
+
   requestConfig = () => {
     this.state.socket.emit('message', {roomId: this.state.roomId, event: 'get config'})
+    setTimeout(() => {
```
rfreitas: you should unlisten to the timeout

rfreitas:
Travis automatic deployment: https://staging-232-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Router/index.js
```diff
@@ -28,23 +30,40 @@ class CrossDeviceMobileRouter extends Component {
...
     this.requestConfig()
   }
 
+  componentDidMount() {
+    this.state.socket.on('disconnect', this.setError)
+  }
+
   requestConfig = () => {
     this.state.socket.emit('message', {roomId: this.state.roomId, event: 'get config'})
+    if (timeoutId) {
```
rfreitas: where is timeoutId coming from here?

src/components/crossDevice/CrossDeviceLink/index.js
```diff
@@ -10,7 +10,8 @@ import { trackComponent } from '../../../Tracker'
 class CrossDeviceLink extends Component {
   constructor(props) {
     super(props)
-    this.state = {roomId: null}
+    const roomId = this.props.roomId || null
+    this.state = {roomId}
```
hjheath: Do we even need roomId in state?

src/components/crossDevice/CrossDeviceLink/index.js
```diff
@@ -47,19 +48,22 @@ class CrossDeviceLink extends Component {
...
     socket.on('clientSuccess', this.onClientSuccess)
-    socket.emit('join', {})
+    socket.emit('join', {roomId: this.state.roomId})
   }
 
   onJoined = (data) => {
-    this.setState({roomId: data.roomId})
+    const {actions, roomId} = this.props
+    if (!roomId) {
+      actions.setRoomId(data.roomId)
```
hjheath: Nice, this is much better in `onJoined` than in `onGetConfig`!

src/components/crossDevice/GenericError/index.js
```diff
@@ -0,0 +1,22 @@
...
+import style from './style.css'
+import { sendScreen } from '../../../Tracker'
+
+class GenericError extends Component {
+  componentDidMount() {
+    sendScreen(['generic_client_error'])
+  }
+  render () {
+    return (
+      <div className={style.base}>
```
hjheath: I think this should use `theme.step` instead

src/components/utils/jwt.js
```diff
@@ -0,0 +1,12 @@
+const parseJwt = (token) => {
+  const base64Url = token.split('.')[1]
+  const base64 = base64Url.replace('-', '+').replace('_', '/')
+  return JSON.parse(atob(base64))
+}
+
+export const jwtExpired = (token) => {
+  if (!token) return true
```
hjheath: I would rather we did this check outside this function because the jwtExpired should not be responsible for whether or not the payload contained a token.

hjheath:
Looks to be working well. What is the expected behaviour if I go back to the upload screen on desktop while connected to mobile?

src/components/Router/index.js
```diff
@@ -28,23 +30,40 @@ class CrossDeviceMobileRouter extends Component {
...
+      error: false,
+      loading: true
     }
     this.state.socket.on('config', this.setConfig(props.actions))
     this.state.socket.emit('join', {roomId: this.state.roomId})
     this.requestConfig()
   }
 
+  componentDidMount() {
+    this.state.socket.on('disconnect', this.setError)
```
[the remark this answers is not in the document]
hjheath: So this covers desktop disconnecting. However if you open two mobile clients and one of them disconnects (or even just refreshes) then you will get the error screen on the other mobile client. Is that OK? @rfreitas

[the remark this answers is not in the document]
hjheath: As agreed it's out of scope. Issue here: https://jira.onfido.co.uk/browse/CX-1337

src/components/Router/index.js
```diff
@@ -28,23 +30,40 @@ class CrossDeviceMobileRouter extends Component {
...
+      if (!this.state.token) this.setError()
+    }, 5000);
   }
 
   setConfig = (actions) => (data) => {
     const {token, steps, documentType, step, woopraCookie} = data
     setWoopraCookie(woopraCookie)
-    this.setState({token, steps, step})
+    if (jwtExpired(token)) this.setError()
+    this.setState({token, steps, step, loading: false})
```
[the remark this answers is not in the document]
stefaniacardenas: Yes, I will add a return.

src/components/crossDevice/GenericError/index.js
stefaniacardenas: True!

src/components/crossDevice/CrossDeviceLink/index.js
stefaniacardenas: Not really

rfreitas:
Travis automatic deployment: https://staging-232-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-232-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Router/index.js
```diff
@@ -28,39 +30,67 @@ class CrossDeviceMobileRouter extends Component {
...
+    if (this.configTimeoutId) {
+      clearTimeout(this.configTimeoutId)
+    }
   }
 
   setConfig = (actions) => (data) => {
     const {token, steps, documentType, step, woopraCookie} = data
     setWoopraCookie(woopraCookie)
-    this.setState({token, steps, step})
+    if (!token || jwtExpired(token)) return this.setError()
```
hjheath: 👍

src/components/Router/index.js
```diff
@@ -28,39 +30,67 @@ class CrossDeviceMobileRouter extends Component {
...
+  setError = () => {
+    this.setState({error: true, loading: false})
+  }
+
   onStepChange = ({step}) => {
     this.setState({step})
   }
 
   sendClientSuccess = () => {
+    this.state.socket.off('disconnect', this.setError)
```
hjheath: I think you should turn off the listener when we unmount as well

src/components/crossDevice/MobileFlow.js
```diff
@@ -12,6 +12,8 @@ class MobileFlow extends Component {
   componentWillUnmount() {
     this.props.socket.off('get config')
     this.props.socket.off('clientSuccess')
+    const {socket, roomId} = this.props
+    socket.emit('disconnecting', {roomId})
```
hjheath: Is this a hack? Why doesn't the socket connection get removed when we unmount?

rfreitas:
Travis automatic deployment: https://staging-232-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-232-pr-onfido-sdk-ui-onfido.surge.sh

src/index.js
```diff
@@ -96,6 +96,8 @@ Onfido.init = (opts) => {
     },
 
     tearDown() {
+      const state = store.getState()
```
rfreitas: not going through a selector feels like bad practice. you should get someone else's perspective on it, because I might not be correct.
However, there are benefits with organizing our reads into queries (ask selectors)

TheComfyChair: The root issue here is that redux is being misused a little. The socket should not be stored in the Redux store as it is cannot be serialized - it's a socket instance. The socket instance should not be stored in the store, and instead should be contained elsewhere - most likely a closure created by a thunk (or saga if you were to be using those). This functionality would handle information coming in from the socket and dispatch events accordingly to keep your store updated. The ability to close this socket would be returned as part of the initialization (which would most likely occur when mounting a certain component), stored on the component instance, and called when unmounting.

Doing things this way keeps the store doing what it's supposed to do - storing data :) Storing implementation details of how to store data directly into the store isn't ideal. If nothing else, you're allowing the entire application to mess with that socket!

rfreitas: nice feedback!
And your perspective of accessing data from the store without using the selector? (imagine we are talking about serializable data)

rfreitas:
Travis automatic deployment: https://staging-232-pr-onfido-sdk-ui-onfido.surge.sh

PR #262, opened 2017-11-20
Spike/cross device tests

hjheath:
PR to check multi-tab tests work on CI

There are a few moving parts here:

- Browser
- Web driver
- Selenium client

I had to refactor the textarea element area so that selenium could pull the text out of it.

I have got a basic scenario of the cross device flow working between two tabs. This works locally and on CI. 

As far as I can tell there is nothing specific to chrome. However I have been unable to test this approach on firefox because my firefox webdriver crashes (for unrelated reasons).

I will not merge this PR, but use it as the starting point for writing the cross device tests.

### DIFF
--- a/src/components/crossDevice/CrossDeviceLink/index.js
+++ b/src/components/crossDevice/CrossDeviceLink/index.js
@@ -110,7 +110,7 @@ class CrossDeviceLinkUI extends Component {
 
   copyToClipboard = (e) => {
     e.preventDefault()
-    this.textArea.select()
+    this.linkText.select()
     document.execCommand('copy')
     e.target.focus()
     this.setState({copySuccess: true})
@@ -217,7 +217,9 @@ class CrossDeviceLinkUI extends Component {
           <div className={style.copyLinkSection}>
             <div className={`${style.label}`}>Copy link instead:</div>
               <div className={classNames(style.actionContainer, {[style.copySuccess]: this.state.copySuccess})}>
-                <textarea className={style.linkText} ref={(textarea) => this.textArea = textarea} value={mobileUrl} />
+                <textarea className={style.linkText} ref={(element) => this.linkText = element}>
+                  {mobileUrl}
+                </textarea>
                 { document.queryCommandSupported('copy') &&
                   <a href='' className={style.copyToClipboard} onClick={this.copyToClipboard}>{linkCopy}</a>
                 }
--- a/test/features/step_definitions/sdk_steps.rb
+++ b/test/features/step_definitions/sdk_steps.rb
@@ -5,6 +5,7 @@ Given(/^I verify with (passport|identity_card|drivers_license)$/) do |document_t
     Then I should see 3 document_select_buttons ()
     When I click on #{document_type} ()
     Then page_title () should contain "Upload front of document"
+    And cross_device_header () should contain "Need to use your mobile to take photos?"
   }
 end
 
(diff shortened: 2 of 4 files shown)


rfreitas:
Travis automatic deployment: https://staging-262-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://262-pr-onfido-sdk-ui-onfido.surge.sh

test/features/step_definitions/sdk_steps.rb
```diff
@@ -17,3 +18,20 @@
     When I click on #{action_button} ()
   }
 end
+
+element = '(\w+(?:|\(.*\)) \(\w*\)(?:| index \d+))'
```
dawiddylowicz: Why is this needed?

dawiddylowicz:
Nicely done!

test/features/step_definitions/sdk_steps.rb
hjheath: It's a regex from the monster gem used to identify HTML elements. It's used in line 24.

For the tests I will move the "open new tab" step to the monster gem so we can reuse the regex there.

dawiddylowicz: Ok I see, thanks

src/components/crossDevice/CrossDeviceLink/index.js
```diff
@@ -217,7 +217,9 @@ class CrossDeviceLinkUI extends Component {
           <div className={style.copyLinkSection}>
             <div className={`${style.label}`}>Copy link instead:</div>
               <div className={classNames(style.actionContainer, {[style.copySuccess]: this.state.copySuccess})}>
-                <textarea className={style.linkText} ref={(textarea) => this.textArea = textarea} value={mobileUrl} />
+                <textarea className={style.linkText} ref={(element) => this.linkText = element}>
```
rfreitas: did anything change?

hjheath: functionally? I don't think so. But the link text appears in the DOM this way so the tests can parse it

rfreitas: saw it now. well, it break the standard I think, because it's a form element. https://stackoverflow.com/questions/5314186/javascript-get-textarea-input-via-value-or-innerhtml

test/features/step_definitions/sdk_steps.rb
```diff
@@ -17,3 +18,20 @@
...
+
+When(/^I open #{element} in a new tab$/) do |element|
+  url = element.text
```
rfreitas: if the element is textarea the url can be retrieved with `element.value`, that way we respect the standard.

src/components/crossDevice/CrossDeviceLink/index.js
hjheath: I can revert this - I found a way to pull the value out using selenium:`element.attribute('value')` rather than `element.text`

This element should probably just be a <p></p>, but I don't plan to change it now

hjheath: http://www.rubydoc.info/gems/selenium-webdriver/0.0.28/Selenium/WebDriver/Element

test/features/step_definitions/sdk_steps.rb
hjheath: I linked an old version of the docs. In the version we are using `element.value` has been removed so we must use `element.attribute('value')`

PR #280, opened 2017-12-05
DO NOT MERGE. Spike: I18n example using Polyglot.js

stefaniacardenas:
## Problem
Testing `Polyglot.js` in order to identify the right library for the internationalisation of the JS SDK

## Exploration
1. Imported `Polyglot.js`
3. Passed a locale as an option at initialisation
4. Created objects with keys/string pairs
2. Followed `Polyglot.js` usage example [here](https://github.com/airbnb/polyglot.js)
3. Return phrase in the selected locale
4. Override default phrases by passing custom strings

### DIFF
--- a/src/components/Router/index.js
+++ b/src/components/Router/index.js
@@ -230,9 +230,11 @@ class HistoryRouter extends Component {
   }
 
   localization = () => {
-    const {locale} = this.props.options
+    const {locale, customStrings} = this.props.options
     const phrases = locales[locale]
-    return new Polyglot({locale, phrases});
+    const polyglot = new Polyglot({locale, phrases});
+    if (customStrings) polyglot.extend(customStrings)
+    return polyglot
   }
 
   componentsList = () => this.buildComponentsList(this.state, this.props)
--- a/src/components/Router/index.js
+++ b/src/components/Router/index.js
@@ -4,6 +4,7 @@ import { connect } from 'react-redux'
 import io from 'socket.io-client'
 import createHistory from 'history/createBrowserHistory'
 import URLSearchParams from 'url-search-params'
+import Polyglot from 'node-polyglot'
 
 import { componentsList } from './StepComponentMap'
 import StepsRouter from './StepsRouter'
@@ -13,6 +14,7 @@ import { unboundActions } from '../../core'
 import { isDesktop } from '../utils'
 import { jwtExpired } from '../utils/jwt'
 import { getWoopraCookie, setWoopraCookie, sendError } from '../../Tracker'
+import locales from '../../../locales'
 
 const history = createHistory()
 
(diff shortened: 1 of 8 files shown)


rfreitas:
Travis automatic deployment: https://release-280-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-280-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://280-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-280-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-280-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://280-pr-onfido-sdk-ui-onfido.surge.sh

hjheath:
Code looks good. The size of the package is ~ 13kb, I thought it was smaller than that?

stefaniacardenas:
My bad. I was checking against `polyglot` instead of `node-polyglot`. Good spot! In that case I will have a look at `I18next` which is 10kb.

stefaniacardenas:
`react-i18next` is 20.6kb and making it work with`i18next`doesn't seem so straight forward. Polyglot seems still the best option so far.

src/index.ejs
```diff
@@ -85,7 +85,9 @@
           onComplete: function() {
             /*callback for when */ console.log("everything is complete")
           },
-          steps: steps
+          steps: steps,
+          locale: 'es',
+          customStrings: {new_string: 'This is my custom string'}
```
rfreitas: it looks like this will be a breaking change. don't think we will be able to preserve the existing partial copy customization

src/components/Router/index.js
```diff
@@ -227,6 +229,14 @@ class HistoryRouter extends Component {
     history.push(path, historyState)
   }
 
+  localization = () => {
```
rfreitas: not sure if I would put it on the router. it's not really routing. maybe where the redux store is being prepped.

rfreitas: also an interesting question to consider if we each component should initialize its own polygot or if it should come from its root

src/index.ejs
stefaniacardenas: Not necessarily. We could still allow partial customisation, like in the welcome screen. We could have two separate components, one with localised strings and one that allows for customisable parts.

src/components/Router/index.js
stefaniacardenas: Yes, I put it in the router for now just to have it working but probably it should be at a lower level.
What would be the benefit of initialising polyglot in each component?

PR #290, opened 2017-12-14
Fix media queries overlap

stefaniacardenas:
This PR fixes the issue with the SDK disappearing if screen width is 480px. This was due to media queries overlap.
I have used `px` instead of `em` because `calc` doesn't work with media queries.

### DIFF
--- a/src/components/Theme/constants.css
+++ b/src/components/Theme/constants.css
@@ -1,5 +1,5 @@
-@custom-media --large (min-width: 30em);
-@custom-media --small-viewport (max-width: 30em);
+@custom-media --large (min-width: 480px);
+@custom-media --small-viewport (max-width: 479px);
 @large-text-margin: 40px;
 @small-text-margin: 24px;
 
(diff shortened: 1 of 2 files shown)


rfreitas:
Travis automatic deployment: https://release-290-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-290-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://290-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Theme/constants.css
```diff
@@ -1,5 +1,5 @@
-@custom-media --large (min-width: 30em);
-@custom-media --small-viewport (max-width: 30em);
+@custom-media --large (min-width: 480px);
+@custom-media --small-viewport (max-width: 479px);
```
stephencookdev: A better solution might be to keep using `em`s, but to avoid mixing `min-width` and `max-width`

You can't have mutually exclusive rules, but just letting rules override each other is fairly standard in CSS.

But if you've been depending on mutually exclusive rules so far, this might cause issues swapping over to it. So just leaving this comment as an fyi

PR #349, opened 2018-06-21
Feature/recovery screen permissions cx 2150

elhesuu:
# Problem
User should see an error/recovery screen instead of the upload UI when permission has not been given

# Solution
Once we know for sure the user blocked camera access (after `getUserMedia` rejects in `react-webcam`) display permissions recovery screen.

## Checklist
_put `n/a` if item is not relevant to PR changes_

- [n/a] Have the CHANGELOG, README, MIGRATION, RELEASE_GUIDELINES docs been updated?
- [n/a] Have new automated tests been implemented?
- [x] Have new manual tests been written down?
- [x] Have tests passed locally?
- [x] Have any new strings been translated?

PR #360, opened 2018-07-03
Feature/x device flow intro cx 2202

elhesuu:
# Problem
- Users without webcam should enter the x-device flow straightaway, with intro screen.
- Users with webcam should see the webcam view straightaway without the file upload or the x-device options

# Solution
By default users without a webcam in the regular flow will be taken into the cross-device flow when trying to upload a face (unless `useWebcam` is explicitly set to `false`, necessary for the `face` step in tests)

If camera presence changes while the user is on the face capture step, they will also be taken to the cross-device flow.

## Checklist

- [x] Have new manual tests been written down?
- [x] Have any new strings been translated?


rfreitas:
Travis automatic deployment: https://staging-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-360-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Camera/CameraTypes.js
```diff
@@ -8,7 +8,7 @@ type CameraCommonType = {
   title: string,
   subTitle: string,
   onUserMedia: Function,
-  onUploadFallback: File => void,
+  onFileUpload: File => void,
```
elhesuu: I renamed this callback to disambiguate what "upload fallback" means

rfreitas:
Travis automatic deployment: https://360-pr-onfido-sdk-ui-onfido.surge.sh

elhesuu:
I tried closing and reopening the PR to stop codeclimate from failing but it didn't work.

src/components/Confirm/index.js
```diff
@@ -234,6 +234,6 @@ const FaceConfirm = (props) =>
   <MapConfirm {...props} method= 'face' />
 
 const DocumentFrontConfirm = appendToTracking(DocumentFrontWrapper, 'front')
-const DocumentBackConfrim = appendToTracking(DocumentBackWrapper, 'back')
+const DocumentBackConfirm = appendToTracking(DocumentBackWrapper, 'back')
```
stefaniacardenas: The old string and the new string look identical to me... 🤔

src/components/crossDevice/Intro/index.js
```diff
@@ -0,0 +1,46 @@
+import { h } from 'preact'
+import classNames from 'classnames'
+import theme from '../../Theme/style.css'
+import style from './style.css'
+import Title from '../../Title'
+import { trackComponent } from '../../../Tracker'
+import {preventDefaultOnClick} from '../../utils'
+
+const Intro = ({i18n, nextStep, componentsList, crossDeviceInitialStepType}) => {
```
stefaniacardenas: You are not calling `componentsList`

src/components/Confirm/index.js
stefaniacardenas: I know they aren't identical cause github can detect the change. I was just wondering what the difference is because I can't see it 😄

elhesuu: DocumentBackCon*fri*m vs DocumentBackCon*fir*m 😄

stefaniacardenas: Ahah, I can see it now!

src/components/Router/index.js
```diff
@@ -155,8 +155,13 @@ class MainRouter extends Component {
     return {steps, token, language, documentType, step: this.state.crossDeviceInitialStep, woopraCookie}
   }
 
-  onFlowChange = (newFlow, newStep, previousFlow, previousStep) => {
-    if (newFlow === "crossDeviceSteps") this.setState({crossDeviceInitialStep: previousStep})
+  onFlowChange = (newFlow, newStep, previousFlow, previousStep, previousStepType) => {
+    if (newFlow === "crossDeviceSteps") {
+      this.setState({
+        crossDeviceInitialStep: previousStep,
+        crossDeviceInitialStepType: previousStepType,
```
rfreitas: what is the `previousStepType`?

rfreitas: if we can easily access it and we have a function for it, might be better not to store in the state, since it's one less thing on our state.

rfreitas:
Travis automatic deployment: https://staging-360-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Theme/style.css
```diff
@@ -32,6 +29,11 @@
 .content {
   flex: 1 0 auto;
   position: relative;
+  display: flex;
+
+  > * {
+    flex-grow: 1;
```
rfreitas: is it safe to apply this to all of the children? Was it the goal of the instruction?

src/components/crossDevice/Intro/index.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+  return (
+    <div className={classNames(theme.fullHeightContainer, style.container)}>
+      <Title
+        title={i18n.t(`cross_device.intro.${ isFace ? 'face' : 'document' }.title`)}
+        subTitle={i18n.t('cross_device.intro.sub_title')}
+      />
+      <div className={classNames(theme.thickWrapper, style.content)}>
+      {
+        Object.keys(steps).map(key =>
```
rfreitas: Minor: are we not using an object mapper some where?

rfreitas:
Travis automatic deployment: https://release-360-pr-onfido-sdk-ui-onfido.surge.sh

src/components/crossDevice/Intro/index.js
```diff
@@ -0,0 +1,46 @@
...
+import classNames from 'classnames'
+import theme from '../../Theme/style.css'
+import style from './style.css'
+import Title from '../../Title'
+import { trackComponent } from '../../../Tracker'
+import {preventDefaultOnClick} from '../../utils'
+
+const Intro = ({i18n, nextStep, componentsList, crossDeviceInitialStepType}) => {
+  const isFace = crossDeviceInitialStepType === 'face'
+  const steps = {
```
rfreitas: since we call so many things steps, maybe `bulletPointsString` might be better?
(I actually don't know if they are bullet points)

src/components/Router/index.js
elhesuu: Definitely not a fan of storing more in state.

When flow changes, `componentsList` is built based on the new flow, so the previous `componentList` gets lost.

But on this new screen we need to know where it's coming from to display different copies, so we either have to:
- recreate the old `componentList` and pass it down somehow to `crossDevice/Intro` and access `oldComponentList[previousStep].type`
- store the current component type in state, in the same way we store `previousStep` and `previousFlow`

src/components/crossDevice/Intro/style.css
```diff
@@ -0,0 +1,55 @@
...
+  flex-grow: 1;
+  justify-content: center;
+}
+
+.step {
+  display: flex;
+  padding-bottom: 32px;
+  position: relative;
+
+  &::before {
```
rfreitas: what will this style?

src/components/Theme/style.css
elhesuu: I don't think so, but would be happy use an object mapper and add it to utils if you think it's useful.

src/components/crossDevice/Intro/index.js
elhesuu: I don't think so, but would be happy use an object mapper and add it to utils if you think it's useful.

src/components/crossDevice/Intro/index.js
elhesuu: They're not, but I agree with renaming it

src/components/crossDevice/Intro/style.css
elhesuu: This is a line that connects all three icons:

image

it needs to stretch from top to bottom, and adjust if copy splits into more/less lines.

src/components/Capture/capture.js
```diff
@@ -124,7 +131,7 @@ class Capture extends Component {
   handleDocument(payload) {
     const documentPayload = this.createDocumentPayload(payload)
     this.createCapture(documentPayload)
-    if (this.props.useWebcam && !this.state.uploadFallback) {
```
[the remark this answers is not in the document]
elhesuu: I did the manual regression test  #2 and ir doesn't seem to fail.

But I think I misunderstood the purpose of `uploadFallback`.

src/components/Router/index.js
rfreitas: hmm, so what is the `previousStep` isn't it the index? If so, at what time can it be used to get its corresponding value?

src/components/crossDevice/Intro/style.css
rfreitas: I see. Might be good to add that as comment otherwise it will be hard to find out from the code.

rfreitas:
Travis automatic deployment: https://staging-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-360-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Theme/style.css
elhesuu: The reason for this change is:
> All actions should sit at the bottom of the box on mobile

We need the content to be of `display: flex` in order to take the full height of its parent. Previously it wasn't `flex`, and in order to take the full height the approach was to absolutely position the child holding the copy. This approach will work, but it won't push siblings to the bottom, causing some overlapping issues like the one described [https://github.com/onfido/onfido-sdk-ui/pull/349#issuecomment-401108985](here).

Also described in #356 and [https://github.com/onfido/onfido-sdk-ui/pull/349#issuecomment-399902699](here).

It should be safe to apply to all first immediate children of `.content` (it's always one):
-by default it will be `flex-direction: row`, and only

src/components/Router/index.js
elhesuu: `previousStep` holds the numeric index of the last step before changing flows. 

Even if we persist that index after changing the flow, we can't extract the step type `("document"|"face")`, because the `componentList` will then change:

Before changing the flow:
```
{component: {...}, {step: {type: "welcome", ...}},
{component: {...}, {step: {type: "document", ...}},
{component: {...}, {step: {type: "document", ...}},
{component: {...}, {step: {type: "document", ...}},
{component: {...}, {step: {type: "face", ...}}],
{component: {...}, {step: {type: "face", ...}}],
{component: {...}, {step: {type: "complete", ...}},
```

After changing the flow:
```
componentList = [
{component: {...}, step: {type: "crossDevice", ...}},
{component: {...}, {step: {type: "crossDevice", ...}},
{component: {...}, {step: {type: "crossDevice", ...}},
{component: {...}, {step: {type: "complete", ...}},
```

Just before changing the flow is the last time we can access `componentList` to inspect the current step's type.

rfreitas:
Travis automatic deployment: https://staging-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-360-pr-onfido-sdk-ui-onfido.surge.sh

seewah:
tested on all browsers - functionally working well. The only thing I have spotted is that the 2nd icon on the selfie x-device intro screen should show a face instead of a document behind the phone. Please see https://onfido.invisionapp.com/d/main/#/console/14394034/306126764/preview. Please update. Thanks

rfreitas:
Travis automatic deployment: https://staging-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://360-pr-onfido-sdk-ui-onfido.surge.sh

seewah:
thanks for updating the icon! however, i spotted another issue. The back button on the selfie x-device intro screen does not work. There does not seem to be any console errors, but the page just does not go back. This is seen on windows + FF/chromes at least.

rfreitas:
Travis automatic deployment: https://release-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://360-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Router/index.js
```diff
@@ -205,11 +204,11 @@ class HistoryRouter extends Component {
 
   initialStep = () => this.state.initialStep === this.state.step && this.state.flow === 'captureSteps'
 
-  changeFlowTo = (newFlow, newStep = 0, entriesBack = 1) => {
+  changeFlowTo = (newFlow, newStep = 0, addStepToHistory) => {
```
rfreitas: which one is called outside of this component, `changeFlowTo` or `setStepIndex`. If it's `changeFlowTo` then the default value for `addStepToHistory` should be set there.

src/components/Router/index.js
```diff
@@ -204,11 +204,11 @@ class HistoryRouter extends Component {
 
   initialStep = () => this.state.initialStep === this.state.step && this.state.flow === 'captureSteps'
 
-  changeFlowTo = (newFlow, newStep=0) => {
+  changeFlowTo = (newFlow, newStep = 0, addStepToHistory) => {
```
stefaniacardenas: We should disable navigation if `addStepToHistory` is `false`. See line 198

elhesuu: Why should we disable navigation?

stefaniacardenas: Because when I open the crossDevice client I can see the back button and when I click on it, it doesn't take me anywhere. We always disable the back button on the first screen.

rfreitas:
Travis automatic deployment: https://release-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://360-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-360-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Router/index.js
```diff
@@ -232,14 +232,18 @@ class HistoryRouter extends Component {
...
+  setStepIndex = (newStepIndex, newFlow, excludeStepFromHistory) => {
     const {flow:currentFlow} = this.state
     const historyState = {
       step: newStepIndex,
       flow: newFlow || currentFlow,
     }
-    const path = `${location.pathname}${location.search}${location.hash}`
-    history.push(path, historyState)
+    if (excludeStepFromHistory) {
+      this.setState(historyState)
```
rfreitas: is `this.props.onStepChange()` need to be called? I actually don't know what it does, just saw that  `onHistoryChange` calls it

elhesuu: Yes, it does.

The component chosen for display is based on `this.state.flow` and `this.state.step`. So only when those change, the component will change.

rfreitas: so is it being called? it seems that if we bypass the history that never gets called.

elhesuu: It's being called if `excludeStepFromHistory` is set to `true`.

If we don't set `excludeStepFromHistory` (default behaviour), a new `history` entry will be created, and only then the state will be updated ([see here](https://github.com/onfido/onfido-sdk-ui/blob/2a93ba2384b17e1eb5ffc44019eadc05590fc214/src/components/Router/index.js#L189)).

If we do set `excludeStepFromHistory` to `true`, no history will be created but the state will be updated.

rfreitas: Yep, I understood the behaviour from a code point of view. The question was whether we needed to call `onStepChange` if `excludeStepFromHistory==true`. I guess we don't have to.

elhesuu: Oh... right, sorry I misread.

Actually I think you might have a point, instead of just `setState` we should do the same history does, without actually adding an entry, and that includes calling the `onStepChange` callback.

rfreitas: who listens to `onStepChange`?
The PR has been merged, but you can create another PR to fix it, if it's indeed a problem.

elhesuu: `CrossDeviceRouter`. If we go back on desktop, mobile should be notified of the step change

PR #362, opened 2018-07-09
Feature/fix edge video cx 2224

stefaniacardenas:
# Problem
Edge has partial `object-fit` support (for `` not for `<video>`)  therefore the live capture stream is not full screen.

# Solution

Apply browser specific styles to handle missing `object-fit` support.

## Checklist

### DIFF
--- a/src/components/Camera/style.css
+++ b/src/components/Camera/style.css
@@ -104,11 +104,12 @@
 }
 
 /* this will only apply styles to Edge */
-@supports (object-fit: cover) and (-ms-ime-align:auto) {
+@supports (-ms-ime-align:auto) {
   .video {
     width:auto;
     @media (--small-viewport) {
-      /* HACK: this will make the video look centered on small screens */
+      /* HACK: this will make the video (kind of) look centered on small screens in absence of `object-fit:cover`.
+      It might not work on all screen sizes*/
       left: -50%;
     }
   }
--- a/src/components/Camera/style.css
+++ b/src/components/Camera/style.css
@@ -94,6 +94,7 @@
   display: block;
   position: absolute;
   height: 100%;
+  width:100%
   top: 0;
   left: 0;
   margin: auto;
@@ -102,9 +103,6 @@
   /*Mirroring transform: scale(-1, 1); */
 }
 
-@supports (object-fit: cover){
-  .video{height: 100%; width:100%}
-}
 /* this will only apply styles to Edge */
 @supports (object-fit: cover) and (-ms-ime-align:auto) {
   .video {
--- a/src/components/Camera/style.css
+++ b/src/components/Camera/style.css
@@ -94,7 +94,7 @@
   display: block;
   position: absolute;
   height: 100%;
-  width:100%
+  width:100%;
   top: 0;
   left: 0;
   margin: auto;
--- a/src/components/Camera/style.css
+++ b/src/components/Camera/style.css
@@ -105,7 +105,7 @@
 @supports (object-fit: cover){
   .video{height: 100%; width:100%}
 }
-/* this will only apply styles to Edge (and IE)*/
+/* this will only apply styles to Edge */
 @supports (object-fit: cover) and (-ms-ime-align:auto) {
   .video {
     height:100%;
--- a/src/components/Camera/style.css
+++ b/src/components/Camera/style.css
@@ -106,7 +106,6 @@
 /* this will only apply styles to Edge */
 @supports (object-fit: cover) and (-ms-ime-align:auto) {
   .video {
-    height:100%;
     width:auto;
     @media (--small-viewport) {
       /* HACK: this will make the video look centered on small screens */
--- a/src/components/Camera/style.css
+++ b/src/components/Camera/style.css
@@ -93,7 +93,6 @@
 .video {
   display: block;
   position: absolute;
-  width: 100%;
   height: 100%;
   top: 0;
   left: 0;
@@ -101,6 +100,21 @@
   object-fit: cover;
   z-index: 0;
   /*Mirroring transform: scale(-1, 1); */
+}
+
+@supports (object-fit: cover){
+  .video{height: 100%; width:100%}
+}
+/* ie edge only gets the following rule */
+@supports (object-fit: cover) and (-ms-ime-align:auto) {
+  .video {
+    height:100%;
+    width:auto;
+    @media (--small-viewport) {
+      /* HACK: this will make the video look centered on small screens */
+      left: -50%;
+    }
+  }
 }
 
 .uploadFallback {
--- a/src/components/Camera/style.css
+++ b/src/components/Camera/style.css
@@ -105,7 +105,7 @@
 @supports (object-fit: cover){
   .video{height: 100%; width:100%}
 }
-/* ie edge only gets the following rule */
+/* this will only apply styles to Edge (and IE)*/
 @supports (object-fit: cover) and (-ms-ime-align:auto) {
   .video {
     height:100%;
--- a/src/components/Camera/index.js
+++ b/src/components/Camera/index.js
@@ -146,4 +146,3 @@ export default class Camera extends React.Component<CameraType, CameraStateType>
     )
   }
 }
-
--- a/src/components/Capture/capture.js
+++ b/src/components/Capture/capture.js
@@ -237,7 +237,7 @@ const CaptureMode = ({method, documentType, side, useCapture, i18n, liveness, ..
     useCapture ?
       <Camera {...{i18n, method, title, subTitle, liveness, ...other}}/> :
       <Uploader {...{i18n, instructions, parentheses, title, subTitle, ...other}}/>
-  )
+    )
 }
 
 const mapStateToProps = (state, props) => {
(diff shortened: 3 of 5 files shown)


rfreitas:
Travis automatic deployment: https://staging-362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://362-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Camera/style.css
```diff
@@ -103,6 +102,21 @@
...
+  .video{height: 100%; width:100%}
+}
+/* this will only apply styles to Edge (and IE)*/
+@supports (object-fit: cover) and (-ms-ime-align:auto) {
+  .video {
+    height:100%;
+    width:auto;
+    @media (--small-viewport) {
+      /* HACK: this will make the video look centered on small screens */
+      left: -50%;
```
stephencookdev: This seems... Dangerous? What happens on screens just before the `--small-viewport` width is hit

stefaniacardenas: Large screen 

screen shot 2018-07-12 at 10 23 26

Small screen

screen shot 2018-07-12 at 10 22 11

src/components/Camera/style.css
```diff
@@ -103,6 +102,21 @@
   /*Mirroring transform: scale(-1, 1); */
 }
 
+@supports (object-fit: cover){
```
rfreitas: what does supports do?

rfreitas: just read about it

src/components/Camera/style.css
```diff
@@ -103,6 +102,21 @@
...
+  .video{height: 100%; width:100%}
+}
+/* this will only apply styles to Edge (and IE)*/
+@supports (object-fit: cover) and (-ms-ime-align:auto) {
```
rfreitas: should it not be `supports not`? I'm a bit confused

stefaniacardenas: Good spot. I had both queries because I had some properties that were shared between Edge and other browsers but I removed them. I just need `@supports (-ms-ime-align:auto) {...}`

stefaniacardenas: Actually, you are right, it should be `not`.

stefaniacardenas: The reason for having both properties is that Edge has `object-fit: cover` support for `` not for `<video>` while IE11 doesn't have `object-fit` support at all. So this will target Edge only. I will change the comment and the PR description

rfreitas:
Travis automatic deployment: https://staging-362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-362-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Camera/style.css
```diff
@@ -103,6 +102,21 @@
   /*Mirroring transform: scale(-1, 1); */
 }
 
+@supports (object-fit: cover){
+  .video{height: 100%; width:100%}
+}
+/* this will only apply styles to Edge */
+@supports (object-fit: cover) and (-ms-ime-align:auto) {
```
rfreitas: hm, I see, so is the object-fit really necessary? it seems we are just trying to detect if it's IE by checking the `-ms-ime-align` property, right?

rfreitas:
Travis automatic deployment: https://362-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Camera/style.css
rfreitas: shouldn't `object-fit: cover` be inside of this closure?

src/components/Camera/style.css
rfreitas: does IE 11 have the same problem?

stefaniacardenas: We could remove `object-fit: cover` because IE11 doesn't not support `getUserMedia` so we wouldn't see the issue there, but this feature query is specific to target Edge and not IE11.

rfreitas: got it

rfreitas:
Travis automatic deployment: https://staging-362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://362-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Camera/style.css
```diff
@@ -103,6 +103,18 @@
   /*Mirroring transform: scale(-1, 1); */
 }
 
+/* this will only apply styles to Edge */
+@supports (object-fit: cover) and (-ms-ime-align:auto) {
+  .video {
+    height:100%;
```
rfreitas: Minor: redundant, since same property is applied before

src/components/Camera/style.css
```diff
@@ -103,6 +103,18 @@
   /*Mirroring transform: scale(-1, 1); */
 }
 
+/* this will only apply styles to Edge */
+@supports (object-fit: cover) and (-ms-ime-align:auto) {
```
rfreitas: `@supports (object-fit: cover)` is redundant, no?

stefaniacardenas: Well, sort of. We know IE doesn't support `getUserMedia` but should they do in the future, this will still only target Edge.

src/components/Camera/style.css
stefaniacardenas: Yes

src/components/Camera/style.css
stefaniacardenas: Unless they also implement  `object-fit: cover` ;)

rfreitas:
Travis automatic deployment: https://staging-362-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Camera/style.css
stephencookdev: Okay, opening this up in the inspector, I think you're depending on a coincidence here.
`width: auto` makes the video be as wide as its normal resolution (which is normally much wider than the screen).

I guess for your device `left: -50%` must _happen_ to get it to centre. If you go mess around with the screen size a bit, you'll see that you've not actually centred it - you've just offset it.

Unless I'm misunderstanding the scope of the `auto` width, and what the screen resolutions could be 🤔 

Does just going for `width: 100vw` or something not work? It looks like that works, and [Edge does support it](https://caniuse.com/#feat=viewport-units)

stefaniacardenas: Yes, it does support it, but if I use `width: 100vw` I will end up with the same problem we are trying to solve. See screenshot. I know using `left: -50%` is a hack and probably won't work everywhere but I don't know how else to fix this problem. I also tried giving the parent a defined height and width but it didn't make any difference...
screen shot 2018-07-13 at 10 57 06

stephencookdev: Ahh, sorry - I misunderstood.

In that case, why not set the width/height of the video with absolute values, from the JS side using the window width/height (& update on screen size change)?

It's also a hack, but a cleaner hack imo, and would be more functionally accurate

src/components/Camera/style.css
rfreitas: well, Edge is the new IE :P

rfreitas:
Travis automatic deployment: https://staging-362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-362-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://362-pr-onfido-sdk-ui-onfido.surge.sh

PR #363, opened 2018-07-10
Rebuild component list to get previous component type

elhesuu:
- Revert adding `prevStepType` state var
- Rebuild component list in `Intro` to access `step.type` based on previous step index.

Also not perfect, since a child component is accessing a domain far outside its predictable area of knowledge, but at least it doesn't require adding more state to one of its ancestors.


rfreitas:
Travis automatic deployment: https://staging-363-pr-onfido-sdk-ui-onfido.surge.sh

PR #365, opened 2018-07-12
Liveness challenges

elhesuu:
# Problem
Users should be able to see randomly generated challenges. They should be able to click "next" to proceed to the next set of challenges. New challenges should be generated when users enter the face capture screen.

# Solution
- Each time `Video` mounts, load a new set of challenges from the API (hardcoded for now)
- Display one challenge at the time once video starts recording
- When recording is stopped, both `blob` and the used set of challenges will added to the capture payload.

## Checklist

- [x] Have any new strings been translated?


rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Camera/Video.js
```diff
@@ -1,64 +1,186 @@
 // @flow
```
elhesuu: @stefaniacardenas , @rfreitas 

I think this file has grown a bit too much, and it's responsibilities are way outside the scope of simply recording a video.

It feels some of the functionality should be moved outside `/Camera`, with pure video recording kept inside `/Camera` in a `Video` component, and fetching challenges and challenge controls inside a `Liveness` component. But on the other hand `Photo` very similar controls, and almost identical presentation, so I'm not sure if decoupling it like this would keep `Video` and `Photo` consistent.

I tried a few approaches, but somehow video state, recording, controls, and challenges are very closely related, so it's something that still needs work. Your opinions and thoughts will be very welcome.

src/components/Camera/style.css
```diff
@@ -16,63 +16,74 @@
 }
 
 .overlayFullScreen {
-  @media (--small-viewport) {
+  /*@media (--small-viewport) {*/
```
elhesuu: Video capturing now looks fullscreen on desktop, so `isFullscreen` should now affect all viewports, and only small viewports:

image

I know `Photo` will also be affected by this, so for now I commented these media queries out until I can confirm `Photo` still needs to be restrained inside a rectangle on larger viewports.

src/components/utils/onfidoApi.js
```diff
@@ -27,6 +27,18 @@ export const uploadLivePhoto = (data, token, onSuccess, onError) => {
   sendFile(endpoint, data, token, onSuccess, onError)
 }
 
+export const getLivenessChallenges = () => {
+  return Promise.resolve([
```
elhesuu: Hard-coded for now

src/components/Overlay/style.css
```diff
@@ -49,7 +41,34 @@
...
+  @media (--small-viewport) {
+    .circle {
+        width: 70%;
+        border: 2.89px solid rgba(255,255,255,1);
+    }
+  }
+}
+*/
+
+.face {
```
elhesuu: I am trying this SVG mask approach in order to reproduce the oval in the designs. Not sure if this has been tried in the past though, but proposing it just in case.

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Camera/Liveness/Challenge.js
```diff
@@ -0,0 +1,55 @@
+// @flow
+import * as React from 'react'
+import { h } from 'preact'
+import classNames from 'classnames'
+import style from './style.css'
```
rfreitas: we are getting quite deep in our nesting, the style selector generator will not namespace these folders, we are risking collision quite a bit.

src/components/Camera/Liveness/Challenge.js
```diff
@@ -0,0 +1,55 @@
...
+    title={i18n.t('capture.liveness.challenges.move_head', {
+      side: i18n.t(`capture.liveness.challenges.${ side }`),
+    })}
+    renderInstructions={() =>
+      <span className={classNames(style.moveHead, style[`moveHead-${ side}`])} />
+    }
+  />
+)
+
+const Challenge = (props: Props) => functionalSwitch(props.type, {
```
rfreitas: Minor: this component seems to be pure as well, therefore naming might be confusing when comparing with `ChallengePure`

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Camera/Video.js
```diff
@@ -1,64 +1,187 @@
...
-  render() {
     return (
-      <div>
-        <CameraPure {...{...this.props, video: true, webcamRef: (c) => { this.webcam = c }}}/>
-        <CaptureActions {...this.props}
-          btnText={this.buttonText()}
-          handleClick={this.handleVideoClick}
-          btnClass={this.buttonClass()}
-        />
+      <div className={style.livenessCamera}>
```
rfreitas: it might be possible to make this its own component.

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

src/components/utils/onfidoApi.js
rfreitas: I would avoid depending on promises so leave room to remove that dependency in the future and decrease bundle size.

src/locales/en.json
```diff
@@ -235,6 +253,10 @@
     "camera_not_working": {
       "message": "Your camera isn’t working",
       "instruction": "It may be disconnected or not functional. <retry>Try again.</retry>"
+    },
+    "liveness_timeout": {
```
stefaniacardenas: Seems like these strings are missing in `es.json`

src/components/Error/index.js
```diff
@@ -19,8 +20,16 @@ const Error = ({className, error, i18n, smaller, renderMessage = identity, rende
       <p className={style.instruction}>
         {renderInstruction(errorText.instruction)}
       </p>
+      {
+        renderAction ?
```
stefaniacardenas: You could do `{ renderAction && <div className={theme.actions}>{renderAction()}</div>}`

src/components/Camera/index.js
```diff
@@ -31,14 +31,14 @@ const UploadFallback = ({onUploadFallback, onFallbackClick, method, i18n}) => {
...
-export const CaptureActions = ({handleClick, btnText, isFullScreen, btnClass}: CameraActionType) => {
+export const CaptureActions = ({children, hint}: CameraActionType) => {
   return (
     <div className={style.captureActions}>
-      <button
-        className={classNames(style.btn, btnClass)}
-        onClick={handleClick}>
-        <div className={classNames({[style.btnText]: isFullScreen})}>{btnText}</div>
-      </button>
+      { hint ?
```
stefaniacardenas: Same here, could be a one liner :)

src/components/Camera/Video.js
stefaniacardenas: Yes, I agree that some functionalities should be moved outside Camera. You could have a wrapper for `Video` that handles recording and challenges. I think we can refactor this logic as part of another story. What do you think @rfreitas?

src/components/Camera/Video.js
```diff
@@ -1,64 +1,187 @@
...
       </div>
     )
   }
+
+  render() {
+    const { i18n } = this.props
+    const { hasSeenIntro, hasLoaded, hasError } = this.state
+
+    return (
+      hasSeenIntro ?
```
stefaniacardenas: Is the user supposed to see the intro again if they navigate back?

src/components/Error/index.js
elhesuu: Not a big fan of short-circuiting! But I can change it ;)

src/components/Camera/Video.js
elhesuu: I assumed they don't. @seewah?

src/components/Camera/Liveness/Challenge.js
elhesuu: Agreed, will rename it

src/components/Camera/Video.js
rfreitas: I'm pretty sure I replied to this.... Github reviews has very poor UX.

anyway. I agree, however I think the main separation has been done already between camera (react-webcam) and business logic (face, liveness, etc). I think for this component we might just have to rename it (LivenessCapture?) and move it to its own folder.

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Camera/style.css
```diff
@@ -8,70 +8,113 @@
...
+    width: 8px;
+    height: 8px;
+    margin-right: 5px;
+    background: #DC2A2A;
+    border-radius: 4px;
+    line-height: 120%;
   }
 }
 
+.iconBtn() {
```
rfreitas: I think multiple selectors is better than mixins. This will repeat the css, won't it? We need to strive for the smallest css.

EDIT. SASS has "virtual" inheritance, but I don't think less has it.

src/components/Camera/style.css
```diff
@@ -176,3 +219,46 @@
   text-decoration: underline;
   cursor: pointer;
 }
+
+.absoluteBottom() {
```
rfreitas: `absoluteBottom` seems misleadingly general, this set of rules is really only fit for `.captureActions`
Is it not possible to put inside of `.captureActions`> maybe with multiple selectors as well?

src/components/Camera/style.css
```diff
@@ -176,3 +219,46 @@
...
+  @media (--small-viewport) {
+    .absoluteBottom();
+  }
+
+  .livenessCamera & {
+    .absoluteBottom();
+  }
+}
+
+.whiteText() {
```
rfreitas: again, the name is very general, but it's applicable to only one element

src/components/Camera/style.css
```diff
@@ -176,3 +219,46 @@
...
+  color: #fff;
+  font-size: 14px;
+  margin-bottom: 10px;
+}
+
+.captureActionsHint {
+  @media (--small-viewport) {
```
rfreitas: why two selectors `@media (--small-viewport)` and `.livenessCamera & ` if the content is the same?

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

src/components/NavigationBar/index.js
```diff
@@ -3,8 +3,11 @@ import classNames from 'classnames'
 import style from './style.css'
 import {preventDefaultOnClick} from '../utils'
 
-const NavigationBar = ({back, i18n, disabled, isFullScreen, className}) =>
-  <div className={classNames(className, style.navigation, {[style.fullScreenNav]: isFullScreen})}>
+const NavigationBar = ({back, i18n, disabled, isFullScreen, isLiveness, className}) =>
+  <div className={classNames(className, style.navigation, {
+    [style.fullScreenNav]: isFullScreen,
+    [style.livenessNav]: isLiveness,
```
rfreitas: hmm, can the liveness screen just use the `isFullScreen` considering it has the same style? Looking at the css it seems difference is on desktop and mobile?

The problem I have more is that the navigation bar is aware of something like `Liveness`. One should be able to control different modes of the navigation bar, but I don't think the navigation bar should be aware of different components.

src/components/Router/StepsRouter.js
```diff
@@ -21,20 +24,26 @@ class StepsRouter extends Component {
...
+    isLiveness,
+  })
 
   render = ({options: {...globalUserOptions}, ...otherProps}) => {
     const componentBlob = this.currentComponent()
     const CurrentComponent = componentBlob.component
     const {back, i18n, disableNavigation} = this.props
     const options = componentBlob.step.options
-    const isFullScreen = this.state.fullScreen
+    const {isFullScreen, isLiveness} = this.state
```
rfreitas: again here, I would avoid having the router be aware of `Liveness`

src/components/NavigationBar/index.js
elhesuu: Couldn't agree more. 

There is a problem with how it currently works, however.  Children components can change the fullscreen state via `useFullScreen`. 

Currently, when fullscreen is enabled, on mobile components will look fullscreen, but on desktop they will look the same.  

My initial work on this PR removed the exception of fullscreen not working on desktop, as Liveness does have to look fullscreen on desktop according to designs. 

The problem is `Photo` still doesn't have too look fullscreen on desktop. So the exception of desktop not showing as fullscreen needs to remain, except for Liveness where we do have to show in fullscreen mode on desktop too.

So now we not only need to set `fullscreen` to `true`, the knowledge of whether is `Liveness` or not is relevant for the whole app, including `Overlay`, `NavigationBar` and a few other components higher up in the component tree.

src/components/Camera/style.css
elhesuu: Same reason discussed [here](https://github.com/onfido/onfido-sdk-ui/pull/365#discussion_r203092465).

src/components/Camera/style.css
elhesuu: It will repeat the CSS, but I see no other way. 

Outer selectors are in scope inside `@media` in less, so `:extend`ing is not an option here 😢

src/components/Camera/style.css
elhesuu: The fact that these selectors are marked with `()` will exclude them from final ouptut. 

The names used for those mixins tried to closely represent to what their real implementation is, and decouple it to what elements they will affect.

elhesuu:
@rfreitas, all CSS issues discussed here stem from the fact there is potentially undesired complexity introduced in the rule that `fullscreen` might or might not really be `fullscreen`, depending on which screen you're seeing and which device you're on.

Maybe this needs discussion from a UI perspective, rather than from an implementation perspective.

src/components/NavigationBar/index.js
elhesuu: @rfreitas, maybe we can rename `isLiveness` to `isFullscreenDesktop`?

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Camera/style.css
```diff
@@ -36,42 +42,76 @@
...
+    width: 8px;
+    height: 8px;
+    margin-right: 5px;
+    background: #DC2A2A;
+    border-radius: 4px;
+    line-height: 120%;
   }
 }
 
+.fullScreenRoundedIcon() {
```
rfreitas: this repetition can be avoided by doing:
```
.fullScreenStartRecordingIcon, . fullScreenStopRecordingIcon {
 (...)
}

.fullScreenStartRecordingIcon() {
   (...)
}

.fullScreenStopRecordingIcon() {
   (...)
}
```

src/components/Camera/style.css
```diff
@@ -36,42 +42,76 @@
...
+  .fullScreenRoundedIcon();
+  background-image: url('./assets/start.svg');
+}
+
+.fullScreenStopRecordingIcon() {
+  .fullScreenRoundedIcon();
+  background-image: url('./assets/stop.svg');
+}
+
 .startRecording {
```
rfreitas: isn't `.startRecording` only active for liveness, if so do you know why we need the small media query?

rfreitas:
There is a visible switch of the mirroring of the webcam, I'm not sure if this was there before.

src/components/Camera/Video.js
rfreitas: I was reading the come more and I'm more convinced that we should move this to its own folder (or to `/Liveness`) and call is `LivenessCapture`

src/components/Liveness/index.js
```diff
@@ -0,0 +1,83 @@
...
+            <Video {...{
+              ...this.props,
+              challenges,
+              onRedo: this.loadChallenges,
+              timeoutSeconds: 15,
+            }} /> 
+          :
+            <Spinner />
+        :
+        <Intro {...{ i18n, onNext: this.handleIntroNext }} />
```
rfreitas: I think this means that the intro screen is not back navigable, right? I don't think we want that.

src/components/Camera/index.js
```diff
@@ -134,7 +132,7 @@ export default class Camera extends React.Component<CameraType, CameraStateType>
     };
     if (this.props.autoCapture) return <AutoCapture {...props} />
     return process.env.LIVENESS_ENABLED && this.props.liveness ?
-      <Video {...props} /> :
+      <Liveness {...props} /> :
```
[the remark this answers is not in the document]
rfreitas: as discussed with @elhesuu because of the Capture component constraint, we can't make the Liveness component own the Camera

src/components/Camera/style.css
rfreitas: so this is not possible because a selector with media query is using these values as well

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

elhesuu:
@rfreitas as discussed, I renamed the component and moved intro as a step. I had to make a few amendments to not show the liveness intro screen if `liveness` not enabled as an option. This also prevents tests from breaking.

I also realised the "redo" button required a few more changes too.

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Camera/LivenessCamera.js
```diff
@@ -0,0 +1,152 @@
...
+  isRecording: boolean,
+};
+
+const initialState = {
+  currentIndex: 0,
+  isRecording: false,
+  hasTimedOut: false,
+}
+
+export default class Video extends React.Component<Props, State> {
```
rfreitas: Minor: this class is still called `Video`

rfreitas: it seems it doesn't need a name anyway

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Camera/index.js
```diff
@@ -65,23 +63,23 @@ export class CameraPure extends React.Component<CameraPureType> {
   }
 
   render() {
-    const {method, title, subTitle, onUploadFallback, onFallbackClick, hasError,
-      onUserMedia, onFailure, webcamRef, isFullScreen, i18n, video} = this.props;
+    const {method, title, subTitle, onUploadFallback, onFallbackClick, error, errorAction, className,
+      onUserMedia, onFailure, webcamRef, isFullScreen, isLiveness, i18n, video} = this.props;
```
rfreitas: I think this `isLiveness` can also be renamed `isFullScreenDesktop`

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Camera/index.js
```diff
@@ -65,23 +63,23 @@ export class CameraPure extends React.Component<CameraPureType> {
...
+      onUserMedia, onFailure, webcamRef, isFullScreen, isFullScreenDesktop, i18n, video} = this.props;
     return (
-      <div className={style.camera}>
-        <Title {...{title, subTitle, isFullScreen}} smaller={true}/>
+      <div className={classNames(style.camera, className)}>
+        <Title {...{title, subTitle, isFullScreen, isFullScreenDesktop}} smaller={true}/>
         <div className={classNames(style["video-overlay"], {[style.overlayFullScreen]: isFullScreen})}>
           {
-            hasError ?
+            error ?
```
stefaniacardenas: We are passing a prop down from the `Capture` parent component called `error`. This means that with this change we will now display any error passed down from the parent, even backend ones. This will interfere with the current autocapture behaviour where we keep sending frames to the document detect service until it does not return a validation error. `hasError` was only checking for camera errors, not for backend ones and I think we should preserve that behavior. Maybe you can rename it to `hasCameraError`.

elhesuu: While completely agreeing with the need for renaming this, in my opinion is a symptom of a bigger problem, that is how props travelling far down the tree are exposed to intermediate components leaving the door open to:
- obscuring data to children by mistake (what I've done now)
- behave in certain ways depending on information not originally intended for their own consumption
- breaking encapsulation by implicit knowledge of responsibilities of other components 

Not intending any actions from the point I'm trying to make, just expressing a thought.

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Camera/index.js
stefaniacardenas: I totally agree, we should not pass props that are not intended for a specific child component. We should avoid that pattern and extend the use of `flow` to every component.

src/components/Camera/index.js
```diff
@@ -50,14 +47,14 @@ const reload = () => window.location.reload()
 // support the precise resolution.
 
 export const CameraPure = ({method, title, subTitle, onUploadFallback,
-                            onFallbackClick, hasError, onUserMedia,
+                            onFallbackClick, hasCameraError, onUserMedia,
                             onFailure, webcamRef, isFullScreen, i18n,
-                            video}: CameraPureType) => (
-  <div className={style.camera}>
-    <Title {...{title, subTitle, isFullScreen}} smaller={true}/>
```
stefaniacardenas: Why are you not passing `isFullScreen` anymore? I think we still need this prop in the `Title` component.

elhesuu: Thanks @stefaniacardenas, I accidentally removed it while fixing conflicts.

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

seewah:
some UI issues spotted:

- liveness intro 
-- desktop: margin should 64px above and below the title (Let's make sure...)
-- mobile: margin should 64px above and 32px below the title (Let's make sure...)
-- spacing between the two bullet points should be 24px instead of 32px
-- continue button to anchor to the button (@stefaniacardenas mentioned there is a standard layout for this?)
-- according to the design, "2 simple actions" and "speaking out loud" should be semi-bolded (but I guess this complicate translation?)

- liveness pre-record
-- "position your face" text should be 64px from the top
-- if screen width is > 360px, there should be "Press record and follow the instructions" line above the record button
-- the oval should be 16px below "position your face" (to leave room for the instruction line mentioned above)

- liveness recording
-- "recording" should be 64px from the top
-- spacing between the recording and instruction should be 16px
-- spacing between instruction and arrow container should be 16px
-- spacing between instruction and numbers container should be 8px
-- oval border should be opacity 10% of white
-- oval fill should be same opacity and the rest of the screen

- timeout
-- when timeout appears, the background should be grayed out with an overlay of opacity 57% of 1C1F21, and nothing should be clickable underneath this overlay
-- timeout message should be 64px from top (like all the other error messages)
-- desktop: timeout message left/right margin should 40px (while mobile remains as 16px)
-- timer does not get reset at the right point. it should start when record starts
-- e.g. if i click record, wait 20 seconds, dismiss message, do nothing, the message appears after 20 seconds straightaway, whereas the expected behaviour should be that the message will appear after 20 seconds when I CLICK RECORD AGAIN

seewah:
one more comment on the configuration. 

after chatting with @rfreitas , to satisfy the acceptance criteria: "dev should be able to switch on video by specifying variant:video as a face step option" please update
```
type: 'face',
    options: {
      variant: options.liveness === "true" ? 'video' : 'photo',````

Also, note that dev can start face step without specifying any variant, in which case it should default to `photo`

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

elhesuu:
@seewah. I've updated all points you mentioned, with the exception of three points related to errors presentation.

I haven't done so, since #386 (released but not in `development`) contains changes that are needed to make the amendments requested.

@stefaniacardenas, are you planning to merge #386 to `development`?

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

seewah:
great i will review. we are going to merge #386 into `development` today. Will let you know

stefaniacardenas:
@elhesuu `master` has been merged into `development`, therefore all the changes within  #386 are now available in `development`

elhesuu:
Thanks @stefaniacardenas !

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

seewah:
Hi I have reviewed the latest changes with @danielspagnolo In general, and the screens are looking much better, thanks!

here is our remaining feedback

- face oval
-- the oval is too small on tall screens, making it very hard for users to position their face (arm extending). Can we refer to the existing face layout on development and see if the new oval can resize/re-position in the same way?

- recording screens
-- line-height of the "recording" label should be 16px
-- line-height of the recording instructions (e.g. "look over...") should be 24px
-- digit challenges: there should be spaces between digits and the dashes to aid readability
-- digit challenges: line-height of the digits themselves should be 64px
-- once the above line-height has been correctly specified, please remove the top-margin from the digits from 16px to 8px

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

seewah:
looks good. just the final couple of comments:
- in the full-container camera view, to maintain consistent top anchoring of the oval between selfie and livenss screens, please only show "Take a selfie" without "position ..." for the selfie screen. Please use the standard non-bold body text style for "Take a selfie" (sorry I know this is a bit of a scope change), so
-- selfie screen: single title "Take a selfie" in body text style
-- liveness screen: single title "Position your face in the oval" in body text style
- oval sizing
-- we want to ensure a pretty consistent oval size across all screensizes, as that is the size of the user's face on the screen after all, but we would like to accommodate the smaller-screen iphone (e.g. iphone 5 SE)
-- can we for breakpoint width <=320px, resize the oval in such a way that there are 48px left and right margins around the oval? Basically we want to make sure that the bottom of the oval is not overlapping with the button too much

seewah:
umm.. I found another issue when doing cross-browser testing. On IE Edge, the video start button is not clickable at all. I am seeing Unable to get property isTypeSupported of undefined or null reference in raven.js in the console :(

elhesuu:
@seewah, just chatted with @stefaniacardenas and the issue is that `MediaRecorder` is not available on Edge (or Safari).

It seems a workaround for Safari has higher priority, but probably we could use something like (this)[https://github.com/zouhir/mediarecorder-polyfill] for Edge.

On the other hand, it would be good to define fallback behaviour. Any ideas?

stefaniacardenas:
I think as part of this epic and until we find a solution for Safari, we should at least handle the errors for unsupported MediaRecorder.

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

elhesuu:
Thanks for the feedback @seewah.

Regarding:
> in the full-container camera view, to maintain consistent top anchoring of the oval between selfie and livenss screens, please only show "Take a selfie" without "position ..." for the selfie screen. Please use the standard non-bold body text style for "Take a selfie" (sorry I know this is a bit of a scope change), so
> -- selfie screen: single title "Take a selfie" in body text style
> -- liveness screen: single title "Position your face in the oval" in body text style

This requirement means we have to add either an extra prop and behaviour to `Title`, a `noSubTitleAndSmallerTitleOnSmallViewports` prop (a better name than that, of course), or use CSS specificity from outside the component (a bit hacky). After asking @stefaniacardenas, she suggested discussing this more.

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

seewah:
OK i agree we can ignore the edge issue as it is mediarecorder related. There is a ticket for that. I will catch up with @stefaniacardenas re: title

seewah:
@elhesuu I noticed one more thing when testing the responsive layout. Previously: 

"if screen width is >= 360px, there should be "Press record and follow the instructions" line above the record button" - i.e. line not shown if width < 360px

was correctly implemented, but in the latest test link, I don't see this behaviour anymore

see 
screen shot 2018-08-01 at 09 46 04

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

seewah:
ok cool I have seen that the selfie screen is only showing the title on an actual mobile device. so that is good, thanks!

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

seewah:
ok final comment:

when the liveness timeout message shows, there should an overlay on the rest of the screen and the rest of the screen should not be interactable - ie. it blocks the interface, whereas the selfie "having camera poblem" message should not be blocking the interface

rfreitas:
Travis automatic deployment: https://staging-365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://365-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-365-pr-onfido-sdk-ui-onfido.surge.sh

PR #367, opened 2018-07-13
Ensure onStepChange prop gets called when changing step

elhesuu:
...without changing history

# Problem

`onStepChange` should be called, even if history doesn't change.

See discussion:
https://github.com/onfido/onfido-sdk-ui/pull/360#discussion_r202410030

# Solution

## Checklist

### DIFF
--- a/src/components/Router/index.js
+++ b/src/components/Router/index.js
@@ -119,7 +119,7 @@ class CrossDeviceMobileRouter extends Component {
   onDisconnectPong = () =>
     this.clearPingTimeout()
 
-  onStepChange = ({step}) => {
+  onStepChange = step => {
     this.setState({step})
   }
 
@@ -186,8 +186,13 @@ class HistoryRouter extends Component {
     this.setStepIndex(this.state.step, this.state.flow)
   }
 
+  componentWillUpdate(nextProps, nextState) {
+    if (nextState.step !== this.state.step) {
+      this.props.onStepChange(nextState.step);
+    }
+  }
+
   onHistoryChange = ({state:historyState}) => {
-    this.props.onStepChange(historyState)
     this.setState({...historyState})
   }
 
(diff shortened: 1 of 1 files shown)


rfreitas:
Travis automatic deployment: https://staging-367-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-367-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://367-pr-onfido-sdk-ui-onfido.surge.sh

src/components/Router/index.js
```diff
@@ -239,7 +239,7 @@ class HistoryRouter extends Component {
       flow: newFlow || currentFlow,
     }
     if (excludeStepFromHistory) {
-      this.setState(historyState)
+      this.onHistoryChange({ state: historyState })
```
rfreitas: I wonder if we should have some indirection here, to prevent confusion in the future. Basically the the instructions inside of `onHistoryChange` could be inside their own method (maybe called `updateHistoryState()`) and then both here and inside of `onHistoryChange` that new method is called.
My concern is that `onHistoryChange` is reacting to the history stack changing, so in the future we might have instructions inside that method that will work if the history stack had in fact been changed.
What do you think?

elhesuu: I agree that imperative call to onHistoryChange might be a source of confusion. 

However, instead of refactoring that behaviour into proxy method to avoid that confusion, I'm proposing directly setting the state, and call the `onStepChange` callback prop if `state.step` effectively changes. 

This way `onHistoryChange` will only get called if history changes.

rfreitas:
Travis automatic deployment: https://staging-367-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://367-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-367-pr-onfido-sdk-ui-onfido.surge.sh

elhesuu:
No, I wasn't...

PR #377, opened 2018-07-20
facing mode was not being passed

rfreitas:
# Problem

Some devices mobile devices show the back facing camera to the user 

# Solution

Pass the `facingMode` to `react-webcam.
Necessary change to `react-webcam`: https://github.com/onfido/react-webcam/pull/24

# Additional
`CameraPure` was no longer a pure component, therefore it was refactored to reestablish its former purity.

## Checklist

- [n/a ] Have the CHANGELOG, README, MIGRATION, RELEASE_GUIDELINES docs been updated?
- [n/a ] Have new automated tests been implemented?
- [ n/a] Have new manual tests been written down?
- [n/a ] Have tests passed locally?
- [ n/a] Have any new strings been translated?


src/components/Camera/index.js
```diff
@@ -90,6 +90,7 @@ export class CameraPure extends React.Component<CameraPureType> {
             className={style.video}
             audio={!!video}
             height={720}
+            facingMode={'user'}
```
stefaniacardenas: Will this return the front facing camera also for document autocapture? 
On another note, on FF this syntax will allow the user to select the facing mode on gUM. If we want to force the user facing mode we should probably use `{facingMode: {exact: 'user'}}` but we should check if it's supported on all browsers

rfreitas: `exact` will break on desktop I believe, because the facingMode is not defined for most webcams.
The assumption is that if a device only has one camera then it will give whatever camera.

I'm not sure if we should concern ourselves a lot with auto capture to be honest. but happy to discuss it.

rfreitas:
Travis automatic deployment: https://staging-377-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-377-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://377-pr-onfido-sdk-ui-onfido.surge.sh

package.json
```diff
@@ -58,7 +58,7 @@
     "eslint-plugin-react": "^6.3.0",
     "extract-text-webpack-plugin": "^2.1.2",
     "file-loader": "^0.9.0",
-    "flow-bin": "^0.63.1",
+    "flow-bin": "^0.76.0",
```
stefaniacardenas: What's this for?

rfreitas: this flow, the type checker

src/components/Camera/index.js
stefaniacardenas: I think, if we stop being concerned about autocapture then we should remove the feature altogether because adding regressions doesn't seem ideal given that there might be clients using it.

rfreitas: actually auto capture is used on desktop only, no? If so then we want it to be user facing.

rfreitas: I agree we should remove it. question is if it's part of the scope or not. As you pointed out is matter of it being a regression or not, which it might not be.

stefaniacardenas: It should be desktop only but I've just seen it on mobile, which is a regression! And I've managed to reproduce the camera error after permissions granted on Samsung S8.

rfreitas:
Travis automatic deployment: https://staging-377-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-377-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://377-pr-onfido-sdk-ui-onfido.surge.sh

PR #384, opened 2018-07-25
Reactored demo app to react

rfreitas:
# Problem

It was pretty hard to add functionality to the demo app, considering the frameworkless we have.

# Solution

Refactor the functionality to use react. The main advantage will be how easy it will be change the sdk options in runtime.

## Checklist

- [n/a] Have tests passed locally?


rfreitas:
Travis automatic deployment: https://staging-384-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-384-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://384-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-384-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-384-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://384-pr-onfido-sdk-ui-onfido.surge.sh

src/demo/demo.js
```diff
@@ -101,3 +55,80 @@ const getToken = function(onSuccess) {
...
+    }
+  }
+
+  componentWillUnmount () {
+    if (this.state.onfidoSdk) this.state.onfidoSdk.tearDown()
+  }
+
+  initSDK = (options)=> {
+    const onfidoSdk = Onfido.init(options)
+    this.setState({onfidoSdk})
```
stephencookdev: what's your opinion on using react state for this, rather than just a class property?
I generally only use React state when the thing I'm storing could possibly need to trigger a re-render when being set (which isn't the case here)

src/demo/demo.js
```diff
@@ -101,3 +55,80 @@ const getToken = function(onSuccess) {
...
+    isModalOpen: this.state.isModalOpen,
+    language,
+    steps,
+    mobileFlow: !!options.link_id,
+    onModalRequestClose: () => {
+      this.setState({isModalOpen: false})
+    }
+  })
+
+  render = () => (
```
stephencookdev: [minor] you shouldn't really bind things that don't need binding, since there's a small performance impact to it
`render` should generally be safe to leave unbound, as it's a framework method

src/demo/demo.js
```diff
@@ -101,3 +55,80 @@ const getToken = function(onSuccess) {
...
+    <div class="container">
+      { useModal &&
+        <button
+          id="button"
+          onClick={ () => this.setState({isModalOpen: true})}>
+            Verify identity
+        </button>
+      }
+      {options.async === "false" && this.state.token === null?
+        <span/> : <SDK options={this.sdkOptions()}></SDK>
```
stephencookdev: [minor] why `<span/>` rather than `null`?

src/demo/demo.js
rfreitas: to be honest, I don't tend to get to that level of optimisation, I tend to follow what reduces the changes of unsuspecting bugs and if there is indeed a performance issue then I optimise. In this case I think `onfidoSDK` will never have to trigger a render, because I even have `shouldComponentUpdate -> false`, so it really doesn't have to be in state. However, this also means that having it on state won't trigger a render either.

stephencookdev: I don't do it for purely optimisation reasons
In my mind it's an indicator to the reader if this particular variable affects the render or not

src/demo/demo.js
rfreitas: yeah tbh I did it only for style in this case, I don't like `returns` cluttering the code :P But your point is valid, it's less optimised. Do you have numbers on performance impact of an arrow vs function? My guess was that is was a negligible impact.

src/demo/demo.js
rfreitas: no reason, I will change it

src/demo/demo.js
stephencookdev: Oh, almost certainly negligible for most use-cases. [Someone created a JSPerf](https://stackoverflow.com/questions/17638305/why-is-bind-slower-than-a-closure) if you're interested, though - looks to be about 70% slower.

src/demo/demo.js
stephencookdev: No, it's normally fairly easy to spot - the only time the assumption changes is when the variable enters the `render` method

At which point you're writing `this.myVar` in the render method, and you think "Oh, that needs to be stateful"

But yeah, it's a fair point - I'm not sure if more junior developers would also think that. So maybe it's not worth it, for that reason 🤔

src/demo/demo.js
```diff
@@ -101,3 +55,80 @@ const getToken = function(onSuccess) {
...
+  sdkOptions = ()=> ({
+    token: this.state.token,
+    useModal,
+    onComplete: () => {
+      /*callback for when */ console.log("everything is complete")
+    },
+    isModalOpen: this.state.isModalOpen,
+    language,
+    steps,
+    mobileFlow: !!options.link_id,
```
stefaniacardenas: Much cleaner 👍

rfreitas:
Travis automatic deployment: https://staging-384-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-384-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-384-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-384-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://384-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-384-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://384-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-384-pr-onfido-sdk-ui-onfido.surge.sh

PR #430, opened 2018-08-28
Do not merge / Experiment - move i18n context

elhesuu:
# Problem
Very often need to pass down `i18n` to distant components as a prop, even if it's not needed in the intermediary components. We could benefit from initialising it once and consuming it only from those components that actually need it,

# Solution
This is just an experiment to see how this could work using the new React context API.

## Checklist


src/locales/context.js
```diff
@@ -0,0 +1,40 @@
...
+    const t = i18n.t.bind(i18n)
+    return (
+      <LocaleContext.Provider value={{ language, t }}>
+        {children}
+      </LocaleContext.Provider>
+    )
+  }
+}
+
+export const localised = WrappedComponent =>
```
stefaniacardenas: This looks good! Not sure if it's a good practice but I was thinking that we could have all the app's children localised by default and we can pass a variable for disabling localisation for children that don't need it. Just a thought

src/locales/index.js
```diff
@@ -5,6 +5,7 @@ import es from './es.json'
 import enMobile from './mobilePhrases/en.json'
 import esMobile from './mobilePhrases/es.json'
 import { isDesktop, parseTags } from '../components/utils'
+export { localised, LocaleProvider } from './context'
```
stefaniacardenas: Do you need `LocaleProvider` here? It seems you aren't using it anywhere in this file

src/components/Router/index.js
```diff
@@ -131,8 +130,8 @@ class CrossDeviceMobileRouter extends Component {
   }
 
   render = (props) =>
-    this.state.loading ? <WrappedSpinner i18n={this.state.i18n} disableNavigation={true} /> :
-      this.state.crossDeviceError ? <WrappedError i18n={this.state.i18n} disableNavigation={true} /> :
+    this.state.loading ? <WrappedSpinner disableNavigation={true} /> :
+      this.state.crossDeviceError ? <WrappedError disableNavigation={true} /> :
         <HistoryRouter {...props} {...this.state}
```
stefaniacardenas: Don't we need to wrap this component in `LocaleProvider`?

src/locales/index.js
elhesuu: It's `export`ing rather than `import`ing 😛

src/components/Router/index.js
elhesuu: Yes, we do. Thanks 😄

src/locales/context.js
elhesuu: Glad that you like it.

> have all the app's children localised by default
I think this could be a way to do so. Any child could consume `t` if they need to, only by being wrapped by the `localised` HOC. Are you concerned that having to call `localise(Component)` is introducing more clutter?

src/locales/index.js
stefaniacardenas: Oops! I'm so used to see imports at the top and exports at the bottom so I didn't realise :)

src/locales/context.js
stefaniacardenas: Yes, that was my concern. A HOC seems like a good idea to me 👍

elhesuu:
Thanks @rfreitas!
- I agree with `t` being too short, I left its original name (`i18n.t`) thinking it was a way of evidencing the original i18n API was untouched. But I would also prefer to change it to something  like `translate()`.

- Flow works well with higher order components, I added flow typing for HOCs in the remove capture PR: https://github.com/onfido/onfido-sdk-ui/pull/445/files#diff-74d69cd36f0dea6444bb549a38c9d389R16

src/components/Error/index.js
```diff
@@ -3,9 +3,10 @@ import classNames from 'classnames'
 import {errors} from '../strings/errors'
 import style from './style.css'
 import { identity } from 'components/utils/func'
+import { localised } from '../../locales'
 
-const Error = ({className, error, i18n, withArrow, renderMessage = identity, renderInstruction = identity, renderAction}) => {
-  const errorList = errors(i18n)
+const Error = ({className, error, t, withArrow, renderMessage = identity, renderInstruction = identity, renderAction}) => {
+  const errorList = errors(t)
```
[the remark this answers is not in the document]
elhesuu: Definitely, great idea

rfreitas:
Travis automatic deployment: https://staging-430-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-430-pr-onfido-sdk-ui-onfido.surge.sh

PR #475, opened 2018-09-24
Feature/refactor routing

elhesuu:
# Problem
Almost all new futures that add screens, will need to touch and modify the routing of the app. To do so is a pain, because the original routing was designed for a very simple and linear routing

We need to:
- Delegate routing to components who can have their own internal routing
- Preserve an historical structure, so that components can control their own history, but follow a standard
- Make sure each component does not change the URL path when changing history. that only history state is changed
- Make history changes namespaced to eliminate conflicts with host app
- Allow the developer to change the steps at runtime, and allow the history stack to adapt to it (maybe reset?)
- Allow for non linear user flows (such as cross device)
- Allow for dynamic screens (screens that don't always show up in the flow, depend on previous state, think permission screen

# Solution
Proposed solution consists in:
- giving the ability to any component to move the flow linearly by providing `nextStep()` and `prevStep()` via context, without any additional knowledge of its location.
- and use a portal to always render in the same DOM position, avoiding unwanted nesting of DOM elements due to component nesting

The idea is:

```jsx
const MainFlow = () => (
  <Flow name="main">
    <Step>
      <Intro />
    </Step>
    <Step>
      <DocumentFlow />
    </Step>
    <Step>
      <Complete />
    </Step>
  </Flow>
)

const DocumentFlow = () => (
  <Flow name="document">
    <Step>
      <FrontCapture />
    </Step>
    <Step>
      <Confirm />
    </Step>
  </Flow>
)
```

Also, dynamic flows (conditionally rendered) could be implemented like:

```jsx
const FrontCapture = () => (
  <div>
    <div>
      <h4>Need to use your mobile to take photos?</h4>
      <DynamicFlow
        renderButton={ enter =>
          <span onClick={ enter }>
            Securely continue verification on your mobile
          </span>
        }
      >
        <CrossDeviceFlow />
      </DynamicFlow>
    </div>
    <div id="uploadArea">
      <input type="file" />
    </div>
  </div>
)


const CrossDeviceFlow = () => (
  <Flow name="crossdevice">
    <Step>
      <CrossDeviceIntro />
    </Step>
    <Step>
      <CrossDeviceLink />
    </Step>
  </Flow>

)
```

It's a lot like a router, except for:

- Flows can be navigated linearly using `prevStep()` and `nextStep()`. When they reach the end or beginning they'll use injected `prevStep()` and `nextStep()` to continue parent flows.
- They will render a new context each injecting references to its own `nextStep()` and `prevStep()`, giving the option to any consuming child of calling `nextStep()` to navigate it.

So far it is very much work in progress, and rendering does not work well, since React component nesting will render additional DOM elements. 

## Pros of this approach

- Encapsulation: Flows only know about their own linearity 
- Composability: Flows can be nested without any central source of knowledge
- Reusability: Flows can be reused regardless of their context

## Cons of this approach

- Visualising the entire application flow becomes harder, as there is no central source of knowledge
- Flexibility comes at the price that any component in the tree can add new steps to the flow

## Issues

- **Solving this with portals sounds possible in theory, but this hasn't been fully achieved yet. At them moment it does render in a portal, but each flow needs to remove existing DOM elements in `componentWillUpdate`**.  An alternative solution is still unclear.
- At the moment parent rendering seems to trigger children re-rendering :(, potentially causing infinite loops with `componentDidMount`


src/components/Flow/index.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  componentDidMount() {
+    const { history } = this.props
+    this.unlisten = history.listen(this.handleHistoryChange)
+    this.handleHistoryChange(history.location)
+  }
+
+  componentWillUpdate() {
+    this.portal.childNodes.forEach(child => child.remove())
```
elhesuu: This does not look like a good idea

stefaniacardenas:
Closing this PR as more design sessions required

PR #558, opened 2019-01-23
Include the `sms_base_url` in the payload to Telephony

stephencookdev:
# Problem

# Solution

## Checklist


rfreitas:
Travis automatic deployment: https://release-558-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-558-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-558-pr-onfido-sdk-ui-onfido.surge.sh

src/components/crossDevice/CrossDeviceLink/index.js
```diff
@@ -167,12 +172,14 @@ class CrossDeviceLinkUI extends Component {
...
+  baseMobileUrl = () =>
     // This lets us test the cross device flow locally and on surge.
     // We use the same location to test the same bundle as the desktop flow.
     process.env.MOBILE_URL === "/" ?
-      `${window.location.origin}?link_id=${this.linkId}` :
-      `${process.env.MOBILE_URL}/${this.linkId}`
+      `${window.location.origin}?link_id=` :
+      `${process.env.MOBILE_URL}/`
+
+  mobileUrl = () => `${this.baseMobileUrl()}${this.linkId}`
```
rfreitas: If I'm getting it right, we cannot do this. It opens a big hole, where someone could use our name to send people elsewhere. happy to chat about it.

stefaniacardenas: Agreed, I mentioned this to Stephen offline. Apart from the security issues, this will also hit the validations when trying to create the jwt. I don't think we should support this feature in our demo app, IMO the best way to go about this is for the acceptor to send a request via postman to create the jwt with the desired parameter.

stephencookdev:
Yes, Stefania already mentioned the issues. I was just leaving this PR open
while I work on other solutions

I disagree that reviewers using postman is the best solution. If at all
possible, I think the demo app should link correctly to itself, rather than
other instances of itself.

This particular case isn't too bad, if we _really_ can't solve this in a
secure way. But as a rule, apps linking to other versions of themselves is
how you end up changing stuff not realising that somewhere along the way
you suddenly ended up on production. Even if people know not to trust the
SMS messages, it's very easy to forget.

See-Wah and I wasted a half hour or so yesterday because of the SMS link
being wrong 😥

stefaniacardenas:
We can still test all the cross device feature on the surge staging link. If we want to test that the sms is correctly sent we can use the production link, and in this case the sms should show the body that we are sending on a production environment, not a different one. If we want to get to the root of this problem we should probably just enable sms delivery on staging and store the staging surge link on telephony, and only send the pr number. I don't think `sms_basw_url` is the right approach to solve this problem because it was build to solve a different issue.

PR #618, opened 2019-04-10
added onStepChange prop and bound to events

asarenski:
# Problem
When utilizing the SDK, it would be nice to be able to know the step of the UI flow.

# Solution
Added a callback prop for stepChange

NOTE this is a naive solution and I have not been able to bootstrap the app with superstatic. I would like to consult with your team to see if this is an easy feature to add.

## Checklist

### DIFF
--- a/src/components/Router/index.js
+++ b/src/components/Router/index.js
@@ -218,6 +218,7 @@ class HistoryRouter extends Component {
   }
 
   onHistoryChange = ({state:historyState}) => {
+    this.props.options.events.emit('stepChange', historyState);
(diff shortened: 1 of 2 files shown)


seewah:
A comment regarding this PR: an issue with this approach is that user drop-off events when they are on id.onfido.com during the cross-device flow will not be propagated automatically to the host website. So we would need to design a more comprehensive solution, instead of merging this PR.

rfreitas:
Closing it since we won't use this code as is, but we might use it for reference. Thanks!

PR #642, opened 2019-05-02
[Feature] CX-3522 Ensure correct keyboard navigation order

josokinas:
# Problem
In order to have navigation accessible we must ensure order of tabbing to elements is correct, also only tabbing to required elements is available.

# Solution
Mainly two features:
- when step changes, the focus must reset the tabbing state so that it starts from the top
- when image preview is in "full screen" mode buttons underneath ("back" button and content controls) must be removed from DOM so that they're not tab-able

I do also needed to re-focus to content when switching "full screen" mode, this is because on image preview when enlarging the image, a button to close enlargement is the same one that was to enlarge, and is not not being announced by screen reader - having content refocused now allows user to tab to that button.

## Checklist
_put `n/a` if item is not relevant to PR changes_

- [x] Have the CHANGELOG, README, MIGRATION, RELEASE_GUIDELINES docs been updated?
- [`n/a`] Have new automated tests been implemented?
- [`n/a`] Have new manual tests been written down?
- [x] Have tests passed locally?
- [`n/a`] Have any new strings been translated?

PR #660, opened 2019-05-16
Feature/cx 3583 better focus management

stefaniacardenas:
# Problem
When navigating across different screens, we need to reset the focus to ensure a good user experience for accessibility users. This problem was addressed in this PR https://github.com/onfido/onfido-sdk-ui/pull/642 but unfortunately, the change introduced a new issue only reproducible on certain screens. 
The issue can be seen on the confirmation screen. When navigating to the confirmation screen the screen reader, instead of announcing the H1, it reads the whole screen content. This is not an ideal user experience as it can be confusing for the user, as well as time consuming and inconsistent.

# Solution

When navigating to confirmation screen (or to any other screen with this issue) the screen readers should only announce the page change by reading the H1, not the whole SDK content.
In general, the initial screen reader and tab starting focus should aligned, and that should be the title element on each screen.
I achieved this by replacing `autoFocus=true` with the `focus()` call over the component reference.
I also removed the focus at SDK container level.

Bonus:
Replaced `aria-live="polite"` with a conditional empty string for cases when the element does not need to be announced.

## Checklist

### DIFF
--- a/src/components/EnlargedPreview/index.js
+++ b/src/components/EnlargedPreview/index.js
@@ -78,7 +78,7 @@ class EnlargedPreview extends Component<Props, State> {
           ref={node => this.previewContainer = node}
           tabIndex={-1}
           aria-label={altTag}
-          aria-live={isExpanded ? "assertive" : "polite"}
+          aria-live={isExpanded ? "assertive" : ""}
           aria-expanded={isExpanded.toString()}
           role='img'
         >
--- a/src/components/PageTitle/index.js
+++ b/src/components/PageTitle/index.js
@@ -1,23 +1,33 @@
-import { h } from 'preact'
+import { h, Component} from 'preact'
 import { withFullScreenState } from '../FullScreen'
 import style from './style.css'
 import classNames from 'classnames'
 
-const PageTitle = ({title, subTitle, smaller, isFullScreen, className}) =>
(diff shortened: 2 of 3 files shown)


rfreitas:
Travis automatic deployment: https://release-660-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-660-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://660-pr-onfido-sdk-ui-onfido.surge.sh

kopijunkie:
Assume this was also agreed on already, but I think reading the H1 first rather than the "Back" button is a better UX anyway.

rfreitas:
Travis automatic deployment: https://staging-660-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-660-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://660-pr-onfido-sdk-ui-onfido.surge.sh

stefaniacardenas:
Yes, that's the intended behaviour as discussed with @seewah and @rfreitas

PR #751, opened 2019-07-25
Make README description of shape of onComplete() argument more accurate

davidpcaldwell:
# Problem
The `document_back` property in the `onComplete()` callback is not documented.

# Solution
Add it to example argument in the README.

## Checklist

- [X] Have the CHANGELOG, README, MIGRATION, RELEASE_GUIDELINES docs been updated?
- [n/a] Have tests passed locally?

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,11 +147,6 @@ Onfido.init({
     //     "id": "0af77131-fd71-4221-a7c1-781f22aacd01",
     //     "variant": "standard"
     //   }
-    //   "document_back": {
-    //     "id": "6f63bfff-066e-4152-8024-3427c5fbf45d",
-    //     "type": "passport",
-    //     "side": "back"
-    //   },
     // }
     //
     // For two-sided documents like driving licences, the object will also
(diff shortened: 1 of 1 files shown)


seewah:
thanks for the PR. The only problem with this example is that he "back" of a passport does not make much sense (we only take one image, which is the photo page). So we should either change the document type to a card-type document with 2 sides, or leave it as it is.

davidpcaldwell:
Or perhaps provide multiple examples, or somewhere in the description provide information about the `document_back` property?

davidpcaldwell:
Added another commit; see what you think of this wording.

davidpcaldwell:
Whoops, added one more commit to remove some duplication.

davidpcaldwell:
Whoops, one more to replace `passport` with `driving_licence` ...

stefaniacardenas:
Thank you for your contribution @davidpcaldwell. I will merge this PR into a branch so that the CI can run properly and we can make any changes if needed. We will include your contribution in the next release.  Many thanks.

PR #766, opened 2019-08-09
Replace sleep() with Selenium explicit wait

bartzet:
# Problem
We have been waiting for UI elements with `sleep()` method which was not stable and made tests flaky. 

# Solution
Apply [explicit waits](https://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/lib/until.html) for UI elements.

## Checklist


rfreitas:
Travis automatic deployment: https://staging-766-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-766-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://766-pr-onfido-sdk-ui-onfido.surge.sh

test/pageobjects/Confirm.js
```diff
@@ -80,6 +81,10 @@ class Confirm extends BasePage {
     this.uploadedVideo.isDisplayed()
     this.driver.executeScript("arguments[0].play();", this.uploadedVideo)
   }
+
+  async waitForConfirmBtnToBeLocated() {
+    this.driver.wait(until.elementLocated(By.css('.onfido-sdk-ui-Confirm-btn-primary')))
```
stefaniacardenas: I think this could be added in `utils/mochaw`. In that file you can modify the function on line 5 to accept another parameter, for example:

```
const $driver = driver => (selector, wait) => {
  if (wait) {
    return driver.wait(until.elementLocated(By.css(selector)))
  }
  return driver.findElement(By.css(selector))
}
```
Then in this file, line 6, you can just do 

```
shouldWait = true
get confirmBtn() { return this.$('.onfido-sdk-ui-Confirm-btn-primary', this.shouldWait)}

**Comment by rfreitas** [comment]
Travis automatic deployment: https://staging-766-pr-onfido-sdk-ui-onfido.surge.sh

**Comment by rfreitas** [comment]
Travis automatic deployment: https://release-766-pr-onfido-sdk-ui-onfido.surge.sh

**Comment by rfreitas** [comment]
Travis automatic deployment: https://766-pr-onfido-sdk-ui-onfido.surge.sh

**Comment by rfreitas** [comment]
isn't the implicit timeout supposed to fix this problem, do you know why increasing it doesn't work?

**Comment by bartzet** [comment]
> isn't the implicit timeout supposed to fix this problem, do you know why increasing it doesn't work?

Will investigate.

**Comment by bartzet** [comment]
> isn't the implicit timeout supposed to fix this problem, do you know why increasing it doesn't work?
> 
> PS. It might be called also implicit wait

I have done some reading and found this nice explanation.


Implicit Wait | Explicit Wait
-- | --
Implicit Wait time is applied to all the elements in the script | Explicit Wait time is applied only to those elements which are intended by us
In Implicit Wait, we don't need to specify "ExpectedConditions" on the element to be located | In Explicit Wait, we need to specify "ExpectedConditions" on the element to be located
It is recommended to use when the elements are located with the time frame specified in implicit wait | It is recommended to use when the elements are taking a long time to load and also for verifying the property of the element like(visibilityOfElementLocated, elementToBeClickable,elementToBeSelected)

I'm not 100% why the implicitWait didn't fix the problem but we defintely saw the problems with elements not being found. With the explicit waits if we focus our wait methods on particular elements it should help. From a couple of runs, I have already seen the change for better.

**Comment by rfreitas** [comment]
Travis automatic deployment: https://staging-766-pr-onfido-sdk-ui-onfido.surge.sh

**Comment by rfreitas** [comment]
Travis automatic deployment: https://766-pr-onfido-sdk-ui-onfido.surge.sh

**Comment by rfreitas** [comment]
Travis automatic deployment: https://release-766-pr-onfido-sdk-ui-onfido.surge.sh

File: test/pageobjects/Confirm.js (hunk 1 above)
Comment (bartzet) [reply test/pageobjects/Confirm.js]: I tried that approach and I didn't see improvement. I noticed more test failures. This is the implementation on the other branch pointing to this one:

https://github.com/onfido/onfido-sdk-ui/compare/improvement/CX-3979-remove-sleep-to-wait-for-elements...improvement/wait-for-ui-elements-one-wait-method?expand=1

Please take a look and let me know if I missed something.

File: test/pageobjects/CrossDeviceMobileNotificationSent.js [hunk 2]
```diff
@@ -50,6 +51,10 @@ class CrossDeviceMobileNotificationSent extends BasePage {
   async clickResendLink() {
     this.resendLink.click()
   }
+
+  async waitForYourMobilePhoneIconToBeLocated() {
+    this.driver.wait(until.elementLocated(By.css('.onfido-sdk-ui-Theme-icon')))
```
rfreitas: shouldn't this be using `yourMobilePhoneIcon` ? like this is redundant

test/pageobjects/CrossDeviceSubmit.js
```diff
@@ -18,6 +19,11 @@ class CrossDeviceSubmit extends BasePage {
   async clickOnSubmitVerificationButton() {
     this.submitVerificationButton.click()
   }
+
+  async waitForDocumentUploadedMessageToBeLocated() {
+    this.driver.wait(until.elementLocated(By.css('li:nth-child(1) > .onfido-sdk-ui-crossDevice-CrossDeviceSubmit-listText')))
```
rfreitas: same as before, shouldn't this refer to a property of the class?

rfreitas:
> > isn't the implicit timeout supposed to fix this problem, do you know why increasing it doesn't work?
> > PS. It might be called also implicit wait
> 
> I have done some reading and found this nice explanation.
> 
> Implicit Wait	Explicit Wait
> Implicit Wait time is applied to all the elements in the script	Explicit Wait time is applied only to those elements which are intended by us
> In Implicit Wait, we don't need to specify "ExpectedConditions" on the element to be located	In Explicit Wait, we need to specify "ExpectedConditions" on the element to be located
> It is recommended to use when the elements are located with the time frame specified in implicit wait	It is recommended to use when the elements are taking a long time to load and also for verifying the property of the element like(visibilityOfElementLocated, elementToBeClickable,elementToBeSelected)
> I'm not 100% why the implicitWait didn't fix the problem but we defintely saw the problems with elements not being found. With the explicit waits if we focus our wait methods on particular elements it should help. From a couple of runs, I have already seen the change for better.

what is the implicit value we have at the moment? did you try changing it?

test/pageobjects/Confirm.js
stefaniacardenas: I had a look and I think Ricardo is right, we should use implicit wait if possible.
I think you can do something like this when the driver is initialised.
`driver.manage().timeouts().implicitWait(2000)`

test/pageobjects/CrossDeviceMobileNotificationSent.js
bartzet: In the beginning, I tried that 

`this.driver.wait(until.elementLocated(By.css(this.yourMobilePhoneIcon)))`

but Selenium expects String as a value and throws:

`InvalidArgumentError: invalid argument: 'value' must be a string` hence I use the selector directly.

test/pageobjects/Confirm.js
bartzet: We already have this in `main.js` 

```
await driver.manage().setTimeouts({
    implicit: 10000,
    pageLoad: 10000
  })
```

bartzet:
> what is the implicit value we have at the moment? did you try changing it?

It's set to 10 seconds atm. Yes, I tried increasing it to 20 seconds. I don't think it made a difference.

I have tried another approach on the `development` branch we have tests in `modal.js`. I removed `sleep(500)`, ran tests, they failed immediately because the implicitWait didn't wait for `openModal` button at all. I tried again by adding explicit wait to wait for button

```
async waitForModalButton() {
    this.driver.wait(until.elementLocated(By.css('#button')))
  }
``` 
that worked without problems. 

I was reading some posts that's not recommended to mix the waits (last answer: https://stackoverflow.com/questions/29474296/clarification-of-the-cause-of-mixing-implicit-and-explicit-waits-of-selenium-doc?answertab=oldest#tab-top)

but from what I observed using only implicit wait doesn't currently work.

test/pageobjects/CrossDeviceMobileNotificationSent.js
rfreitas: but it's not really a string, because you are passing the output of `By.css(string)`, isn't it?

rfreitas: I checked and the difference is that $ actually returns `driver.findElement(By.css())` instead of just `By.css(selector)`. if you are going to share the selectors I guess you can have it as a property called something like `[elementName]Selector`?

rfreitas: e.g.
`get yourMobilePhoneIconSelector() { return By.css('.onfido-sdk-ui-Theme-icon')}`
`get yourMobilePhoneIcon() { return this.$(this.yourMobilePhoneIconSelector)}`

rfreitas:
Travis automatic deployment: https://staging-766-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-766-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-766-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-766-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://766-pr-onfido-sdk-ui-onfido.surge.sh

test/pageobjects/CrossDeviceMobileNotificationSent.js
bartzet: Changed in 3f9aac2

test/pageobjects/Confirm.js
```diff
@@ -80,6 +82,10 @@ class Confirm extends BasePage {
     this.uploadedVideo.isDisplayed()
     this.driver.executeScript("arguments[0].play();", this.uploadedVideo)
   }
+
+  async waitForConfirmBtnToBeLocated() {
+    this.driver.wait(until.elementLocated(this.confirmBtnSelector))
```
rfreitas: minor, you could also make this into a more concise function that gets reused.
```
async waitForElement(elementCss) {
   this.driver.wait(until.elementLocated(element))
}
```

stefaniacardenas: Yes, I was about to suggest the same. It will avoid repetition

test/specs/scenarios/crossDevice.js
```diff
@@ -47,10 +48,9 @@ export const crossDeviceScenarios = async (lang) => {
     }
 
     const waitForAlertToAppearAndSendSms = async () => {
-      driver.sleep(1000)
+      driver.wait(until.alertIsPresent())
```
stefaniacardenas: What's this?

stefaniacardenas: I guess a built in selenium-webdriver function to check if there is an alert, correct?

bartzet: Correct and it works really fast ⚡️

bartzet: [Official documentation](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/lib/until.js#L107)

rfreitas:
Travis automatic deployment: https://staging-766-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-766-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://766-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-766-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-766-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://766-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-766-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-766-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://766-pr-onfido-sdk-ui-onfido.surge.sh

PR #871, opened 2019-11-26
QR code library POC

kopijunkie:
POC for QR code generation to get a better idea of bundle size impact

### DIFF
--- a/src/components/crossDevice/CrossDeviceLink/index.js
+++ b/src/components/crossDevice/CrossDeviceLink/index.js
@@ -1,5 +1,8 @@
 import { h, Component} from 'preact'
 import classNames from 'classnames'
+// import { QRCode }  from 'react-qr-svg'
+import QRCode from 'qrcode.react'
+// import * as qrcode from 'qrcode-generator'
 
 import theme from '../../Theme/style.css'
 import style from './style.css'
@@ -16,9 +19,9 @@ import { createSocket } from '~utils/crossDeviceSync'
 
 class SmsError extends Component {
   componentDidMount() {
-     const errorName = this.props.error.name.toLowerCase()
-     this.props.trackScreen([errorName])
-   }
+    const errorName = this.props.error.name.toLowerCase()
+    this.props.trackScreen([errorName])
+  }
   render = ({error}) => <Error role="alert" {...{error}} />
 }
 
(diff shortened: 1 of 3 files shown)


rfreitas:
Travis automatic deployment: https://staging-871-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-871-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://871-pr-onfido-sdk-ui-onfido.surge.sh

src/components/crossDevice/CrossDeviceLink/index.js
```diff
@@ -1,5 +1,8 @@
 import { h, Component} from 'preact'
 import classNames from 'classnames'
+// import { QRCode }  from 'react-qr-svg'
+import QRCode from 'qrcode.react'
```
stefaniacardenas: There is a 6% size increase with this package, which is not huge, but I would still consider splitting this code in a separate bundle to limit size impact.

kopijunkie: I was thinking this would be part of the lazy loaded cross device bundle. 
Is there anything I will need to set up in the Webpack config to ensure this?

stefaniacardenas: Currently we only have the number input validation in the cross device chunk. We can do something similar to what we do [here](https://github.com/onfido/onfido-sdk-ui/blob/fb8cc7cd9b3df2667c17aa15b5e64db02fcf49b5/src/components/PhoneNumberInput/Lazy.js#L36) or refactor that code to include QR

kopijunkie: OK. Will add a note about this in the implementation ticket. I think making a separate lazy loaded QR code generator like what's being done for phone number input would be the way to go given how we'll be implementing it, i.e. the relevant UI is displayed based on user selection.

src/components/crossDevice/CrossDeviceLink/index.js
```diff
@@ -248,6 +262,15 @@ class CrossDeviceLinkUI extends Component {
           <div role="alert" aria-atomic="true">
             { invalidNumber && <div className={style.numberError}>{translate('errors.invalid_number.message')}</div> }
           </div>
+          <div>
+            <br />
+            {/* react-qr-svg - generated SVG size: 83.5 KB
+            <QRCode level="Q" style={{ width: 100 }} value={mobileUrl} /> */}
+            {/* qrcode.react - generated SVG size: 3.8 KB */}
+            <QRCode renderAs="svg" level="Q" size={100} value={mobileUrl} />
+            {/* qrcode-generator (HTML placeholder)
```
joninvski: :heart:

PR #930, opened 2020-03-02
PoC /send analytics to onfido cx 4654

stefaniacardenas:
# Problem

Currently, we send analytics directly to Woopra. We want to use an Onfido proxy to send the analytics to our analytics provider.

# Solution
Send analytics to https://analytics-sdk-dev.onfido.com/v1/import. This requires quite a lot of data manipulation because the service has been built to handle data from segment.io, not Woopra.
This approach does not work due to CORS and this PR has been created as PoC only.

## Checklist


rfreitas:
Travis automatic deployment: https://staging-930-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-930-pr-onfido-sdk-ui-onfido.surge.sh

src/Tracker/utils/index.js
```diff
@@ -0,0 +1,45 @@
...
+        },
+        screen: {
+          height: 480, //made up values, it will break without
+          width: 320 //made up values, it will break without
+        },
+        ...woopra.visitorData
+      },
+      event: eventName,
+      properties,
+      timestamp: woopra.last_activity,
```
joninvski: Isn't this weird. Because woopra is not being used, do you think the `last_activity` value is correct?

src/Tracker/utils/index.js
```diff
@@ -0,0 +1,45 @@
...
+  JSON.stringify({
+    batch: [{
+      anonymousId: woopra.cookie,
+      channel: "Web SDK",
+      context: {
+        app: {
+          name: woopra.instanceName,
+          namespace: woopra.options.domain
+        },
+        library: {
```
joninvski: I can make this optional in the backend... Almost certain that i can

src/Tracker/utils/index.js
```diff
@@ -0,0 +1,45 @@
...
+    else {
+      console.log(request)
+    }
+  }
+  request.onerror = () => console.log(request)
+
+  request.send(payload)
+}
+
+export const formatAnalytics = (woopra, eventName, properties) =>
```
joninvski: aren't you missing `referer` like this? And maybe other important values?

src/Tracker/utils/index.js
```diff
@@ -0,0 +1,45 @@
...
+      anonymousId: woopra.cookie,
+      channel: "Web SDK",
+      context: {
+        app: {
+          name: woopra.instanceName,
+          namespace: woopra.options.domain
+        },
+        library: {
+          name: "analytics-js", //made up values, I don't know what this is
+          version: "0.0.0" //made up values, I don't know what this is
```
kopijunkie: Potentially these 2 properties should be our SDK's name and version?

src/Tracker/utils/index.js
```diff
@@ -0,0 +1,45 @@
+export const sendAnalytics = (payload) => {
+  const request = new XMLHttpRequest()
+  request.open('POST', 'https://analytics-sdk-dev.onfido.com/v1/import')
+  request.setRequestHeader('Content-Type', "application/json")
+  request.setRequestHeader('Authorization', "Basic M002Z3tQWURwWTlUanlDKDNRTHIqTDZOQ1hVY2FLZUU6")
```
kopijunkie: Shouldn't this be hidden from the public repo?

stefaniacardenas: It's not a real secret apparently, I've hardcoded values inside this PoC but I will move then in the webpack config file when this is ready

src/Tracker/utils/index.js
stefaniacardenas: Yes, there is some info missing. This is just a basic "working" example (where the service returns 200). I will add the missing info now. Btw, I haven't checked yet if woopra is actually receiving this data from the service cause there were some issues on the woopra dashboard last week.

src/Tracker/utils/index.js
stefaniacardenas: I'm not sure what you mean when you say "woopra is not being used". I'm using Woopra to track these events

src/Tracker/utils/index.js
stefaniacardenas: I think this refers to the library we used to capture the events, so in this case it would be woopra. I'm not sure if we care about capturing these values. The sdk version is captured inside `woopra. visitorData`

rfreitas:
Travis automatic deployment: https://staging-930-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-930-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-930-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-930-pr-onfido-sdk-ui-onfido.surge.sh

src/Tracker/utils/index.js
stefaniacardenas: It looks like it's already optional

rfreitas:
Travis automatic deployment: https://staging-930-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-930-pr-onfido-sdk-ui-onfido.surge.sh

src/Tracker/utils/index.js
joninvski: but you are never calling the `track` method in woopra library. Are you sure the `last_activity` field is being handled even with this limitation?

Seems weird to be relying on a library but then not really using it as we are supposed to.

src/Tracker/utils/index.js
joninvski: > This is just a basic "working" example (where the service returns 200).

Please *do not* trust the 200 as everything going well. Really. Please double check in woopra. Small note, the current backend will use the mobile projects, not the JS..... that is something we also need to solve.

src/Tracker/utils/index.js
joninvski: Agree that this is not a real secret. Everyone using the JS SDK will be able to see it

src/Tracker/utils/index.js
stefaniacardenas: It does work, regardless of the fact that we are not calling track because it's just returning `new Date()` when you call `woopra. last_activity`
https://github.com/Woopra/browser-tracker/blob/master/src/tracker.js#L85

joninvski: note that code is when you first initialize it. Not when you send a new event.

makes sense?

stefaniacardenas: Sorry, wrong line.
https://github.com/Woopra/browser-tracker/blob/75e4357b15ddc5ccf35acd4caec4a6fd17b178e0/src/tracker.js#L180
There is an event listener listening for move movements which updates `last_activity` value

rfreitas:
Travis automatic deployment: https://staging-930-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://release-930-pr-onfido-sdk-ui-onfido.surge.sh

PR #978, opened 2020-04-28
Use dotenv-safe to manage environment variables

markwilson:
# Problem
For the demo app, there's no list of expected/available environment variables. This means it can be unclear when setting it up what needs to be configured.

It also currently expects the environment variables exist already where the run command is executed, which means either the developer will set it in their shell defaults or in the current shell - this _could_ lead to naming clashes with other projects.

# Solution
Introducing `dotenv-safe` means we can explicitly define a sample of environment variables (with sensible defaults and descriptions, if needed). The `.env` file will be picked up by both the `npm run dev` and `docker-compose` configuration.

## Checklist

- ~Has the CHANGELOG been updated?~
- ~Has the README been updated?~
- ~Has the RELEASE_GUIDELINES been updated?~
- ~Has the MIGRATION doc been updated for any MAJOR breaking changes?~
- ~Has the MIGRATION doc been updated for any MINOR breaking changes, including any translation strings or keys changes?~
- ~Have any new automated tests been implemented or the existing ones changed?~
- ~Have any new manual tests been written down or the existing ones changed?~
- ~Have any new strings been translated or the existing ones changed?~


rfreitas:
Travis automatic deployment: https://release-978-pr-onfido-sdk-ui-onfido.surge.sh

rfreitas:
Travis automatic deployment: https://staging-978-pr-onfido-sdk-ui-onfido.surge.sh

package.json
```diff
@@ -142,6 +142,7 @@
     "blueimp-load-image": "~2.12.2",
     "classnames": "~2.2.5",
     "custom-event-polyfill": "^1.0.7",
+    "dotenv-safe": "^8.2.0",
```
stefaniacardenas, line 145: I think this should go in dev dependencies

.env.example
```diff
@@ -0,0 +1 @@
+SDK_TOKEN_FACTORY_SECRET=
```
stefaniacardenas, line 1: If we decide to pull translations from Lokalise at runtime we could later on add `LOKALISE_TOKEN` and `LOKILISE_PROJECT_ID` here

webpack.config.babel.js
```diff
@@ -246,7 +197,7 @@ const configDist = {
     library: 'Onfido',
     libraryTarget: 'umd',
     path: `${__dirname}/dist`,
-    publicPath: CONFIG.PUBLIC_PATH,
+    publicPath: CONFIG.PUBLIC_PATH ?? `${CONFIG.PUBLIC_PATH_PREFIX}/${packageJson.version}/`,
```
stefaniacardenas, line 200: Looks like `PUBLIC_PATH` is defined for all environments as `/` so `PUBLIC_PATH_PREFIX` will never be returned.

PR #1413, opened 2021-04-20
Migrate CI build from TravisCI to Github Actions

stefaniacardenas:
# Problem
TravisCI only allows 3 concurrent jobs and that's not enough for the team's need. Additionally, they have introduced a new free plan where we are completely blocked from running builds when we run out of credits.

# Solution
Use Github Actions as a CI/CD tool.
The `CI.yml` workflow has two jobs: `test` and `production`.

In the `test` job we run the test build and run ui and integration tests. The UI tests are now running in headless mode. This change was made because the UI tests were failing intermittently when not on headless mode.
For future reference, the error seen was `Error executing chrome test case WebDriverError: unknown error: Chrome failed to start: exited abnormally. (unknown error: DevToolsActivePort file doesn't exist)`.

In the `production` job we run the production build and and deploy to surge. This makes the acceptance process much faster. Only one PR link will be generated and there will be no duplicate comments, so you should expect to only see one comment with the surge link per PR even when deploying to surge multiple times.
The staging link won't be generated anymore because requests against Onfido API staging can be made by passing a staging JWT.

Bonus:
* Only run the SVG optimizer action when there is any changes on svg files
* Remove CodeQL action

## Checklist

- [x] Has the CONTRIBUTING doc been updated?


.github/workflows/CI.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Create Surge Domain
+        if: matrix.NODE_ENV == 'production'
+        run: |
+          IFS='/' read -ra REPO_SLUG_ARRAY <<< "$GITHUB_REPOSITORY"
+          REPO_OWNER=${REPO_SLUG_ARRAY[0]}
+          REPO_NAME=${REPO_SLUG_ARRAY[1]}
+          echo "SURGE_DOMAIN=https://${PR_NUMBER}-pr-${REPO_NAME}-${REPO_OWNER}.surge.sh" >> $GITHUB_ENV
+
+      - name: Publish to surge
```
kopijunkie, line 76:
```suggestion
      - name: Publish to Surge
```

.github/workflows/publish_release.yml
```diff
@@ -60,6 +60,14 @@ jobs:
             echo Publishing next tag $VERSION
             npm publish --tag next
           fi
-
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
+
+      - name: Publish to Surge
```
kopijunkie, line 66:
```suggestion
      - name: Publish latest version to Surge
```

package.json
```diff
@@ -18,7 +18,7 @@
     "postbuild": "dts-bundle-generator src/types/index.ts --out-file lib/index.d.ts",
     "mkdirs": "mkdir -p dist && mkdir -p lib",
     "clean": "rimraf lib dist",
-    "prepare": "npm run clean && npm run check && npm run build",
+    "prepare": "npm run clean",
```
ethan605, line 21: Is this a permanent change or just temporarily during development? :thinking:

test/config.json
```diff
@@ -12,7 +12,7 @@
     },
     {
       "file": "test/specs/multipleBrowsersDocumentUpload.js",
-      "timeout": 50000,
+      "timeout": 75000,
```
ethan605, line 15: :open_mouth:

ethan605 (approved):
LGTM, just a minor concern about the npm `prepare` task

src/components/utils/__integrations__/onfidoApi/document.integration.js
```diff
@@ -36,7 +36,7 @@ const TEST_DOCUMENT_DATA = {
 
 describe('API uploadDocument endpoint', () => {
   beforeEach(async () => {
-    jest.setTimeout(10000)
+    jest.setTimeout(15000)
```
kopijunkie, line 39: Interesting that the JWT needs a bit more time when running on Github Action jobs. 🤔

test/config.json
kopijunkie, line 15: Is this timeout adjustment based on the earlier JWT timeout value increase?

test/mock-server/Dockerfile
```diff
@@ -8,6 +8,6 @@ USER deno
 
 COPY *.ts *.pem lock.json /app/
 COPY /frontend/* /app/frontend/
-RUN deno cache deps.ts --lock=lock.json
+RUN deno cache /app/deps.ts --lock=lock.json
```
kopijunkie, line 11: Think I'm missing something here, but why is this change needed?

kopijunkie (changes requested):
Since Travis CI builds & CodeQL workflow are now removed, the [CONTRIBUTING doc](https://github.com/onfido/onfido-sdk-ui/blob/development/CONTRIBUTING.md#L50) needs to be updated for what pull request checks are run (my bad also for forgetting about this for the SVGO workflow when I added it).

src/components/utils/__integrations__/onfidoApi/document.integration.js
stefaniacardenas: Yes, for some reason some times it fails saying that the callback didn't run within the specified timeout

test/config.json
stefaniacardenas: No, this is because I've noticed that sometimes tests on browserstack are successful but they might take longer than a minute (in rare cases it might take even up to 2 mins). This means that our tests will fail while browserstack is still running the tests, So I decided to increase the timeout

test/mock-server/Dockerfile
stefaniacardenas: It was not able to find `deps.ts` without adding without specifying the home directory

ethan605: Actually, either way is correct because of:

```Dockerfile
COPY *.ts *.pem lock.json /app/
```

which copied all `*.ts` files (including `deps.ts`) into `/app`. But because `/app` was already the `WORKDIR`, so it should work even without `/app` in `deps.ts` path.

As you can see in the last line:

```Dockerfile
CMD ["run", "--allow-net", "--allow-read", "--allow-write", "server.ts"]
```

there's no `/app` at all

package.json
stefaniacardenas: Temporary. Brought it back

test/mock-server/Dockerfile
stefaniacardenas: Cool. I've changed it

test/config.json
kopijunkie: Wow! 2 mins is long. But we can always revisit this later if the current timeout is good enough.

kopijunkie (approved):
🚀

test/config.json
stefaniacardenas: Two mins is rare but I've seen 1:13 quite often

PR #1414, opened 2021-04-20
feature/Integrating FaceTec Auth SDK with Web SDK

carlagn:
# Problem
Integrating FaceTec's Authentication SDK flow inside our own Onfido SDK.
Issues PR tackles:

- Sizing differentiation between normal IDV and IDV + Auth builds (Integrators should **not** be impacted in terms of bundle size when using **only** IDV)
- Initialisation of a FaceTecSession for Auth, following and initialising in production mode in communication with our own internal dedicated backend for auth.
- Minor Look & Feel aspect; PR contains an initial draft that tackles **UI/UX** onfido L/F, lokalise translations, Auth flow starting from "Intro" to "Complete" screen. 

# Solution

- Sizing differentiation is tackled exclusively on `webpack.config.babel.js`. Building for Auth is possible by setting `SDK_ENV=auth` or running `npm run build:auth`/`npm run dev:auth`.
- Initialisation of Auth and Auth modules are set under `src/components/Auth`. Image directory resource for FaceTec is contained inside `core-sdk/FaceTec_images`.
- UI/UX and Lokalize changes can be found in `AuthConfig.js` and `FaceTecStrings.js` respectively.

❗ **Relevant notes regarding changes that should ease the reviewing process:**

- **29** out of the 68 file changes reported originate from the `core-sdk` facetec sdk folder. This folder is the core functionality of the Authentication provided from FaceTec's technology. This folder includes wasm, js, and ts files for the development part, plus image and binary resources for the rendering of the screens.
- **3 466** added lines originate from the `package-lock.json` file. This file already takes into account the changes in development, as the PR has been solved for conflicts with `development`. This is due to some `@types` libraries being update, as well as webpack plugins being added in order to process and generate builds in different ways (for idv and idv+auth).
- **2 651** added lines originate from the `core-sdk/` folder. This should **not** need input from reviewers as it comes directly from FaceTec's technology.
- Each translation file has been updated with additional 94 lines.
- I would request particular focus on the files under `src/components/Auth`, as they represent our main structure for delivering Auth to future clients.
- **E2E Tests: https://gitlab.eu-west-1.mgmt.onfido.xyz/onfido/biometrics/auto/facetec_tests/**

## Checklist

- [x] Have any new strings been translated or the existing ones changed?


.eslintignore
```diff
@@ -2,3 +2,4 @@ test/*.conf.js
 node_modules
 test/mock-server/**/*.js
 coverage
+src/components/Auth/assets/animations.scss
```
ethan605, line 5: You can run `prettier --write src/components/Auth/assets/animations.scss` to fix the lint errors

.prettierignore
```diff
@@ -5,3 +5,4 @@
 /src/supported-documents/*.json
 /test/mock-server/frontend
 /coverage
+/core-sdk
```
ethan605, line 8: Should be specific to `/core-sdk/FaceTec` in case of we need to introduce new things in `core-sdk` in the future. But quick question: why it's `core-sdk`? Should it be `auth-sdk` instead?

tsconfig.json
```diff
@@ -7,9 +7,9 @@
     "jsxFactory": "h",
     "jsxFragmentFactory": "Fragment",
     "lib": ["esnext", "dom"],
-    "module": "es6",
+    "module": "esnext",
     "moduleResolution": "node",
-    "noImplicitAny": true,
+    "noImplicitAny": false,
```
ethan605, line 12: This is an anti-pattern. We introduced `noImplicitAny: true` from the beginning and it should be kept as is. Of course, this should trigger some errors in your implementation, but let's aim to fix it instead of turning off this flag.

tsconfig.json
```diff
@@ -7,9 +7,9 @@
     "jsxFactory": "h",
     "jsxFragmentFactory": "Fragment",
     "lib": ["esnext", "dom"],
-    "module": "es6",
+    "module": "esnext",
```
ethan605, line 10: Any specific reason to change from `es6` to `esnext` here?

webpack.config.babel.js
```diff
@@ -122,6 +127,7 @@ const PROD_CONFIG = {
   SMS_DELIVERY_URL: 'https://telephony.onfido.com',
   PUBLIC_PATH: `https://assets.onfido.com/web-sdk-releases/${packageJson.version}/`,
   USER_CONSENT_URL: 'https://assets.onfido.com/consent/user_consent.html',
+  OLD_AUTH: 'https://edge.api.onfido.com/v3',
```
ethan605, line 130: It's a bit vague to introduce a new `OLD_AUTH` here, can you explain the context?

webpack.config.babel.js
```diff
@@ -260,27 +269,26 @@ const baseConfig = {
   devtool: PRODUCTION_BUILD ? 'source-map' : 'eval-cheap-source-map',
 }
 
-const configDist = {
+const configDist = () => ({
   ...baseConfig,
 
   entry: {
-    onfido: './index.tsx',
+    [`onfido${SDK_ENV === 'auth' ? SDK_ENV : ''}`]: './index.tsx',
```
ethan605, line 276: IMO it should be `'Auth'` rather than `'auth'` to form `onfidoAuth` instead of `onfidoauth`.

webpack.config.babel.js
```diff
@@ -260,27 +269,26 @@ const baseConfig = {
...
     demo: './demo/demo.tsx',
     previewer: './demo/previewer.tsx',
   },
 
   output: {
-    library: 'Onfido',
+    library: `Onfido${SDK_ENV === 'auth' ? SDK_ENV : ''}`,
```
ethan605, line 282: Ditto here (`'Auth'` rather than `'auth'`)

webpack.config.babel.js
```diff
@@ -260,27 +269,26 @@ const baseConfig = {
...
     libraryTarget: 'umd',
     path: `${__dirname}/dist`,
     publicPath: CONFIG.PUBLIC_PATH,
     filename: '[name].min.js',
-    chunkFilename: 'onfido.[name].min.js',
+    chunkFilename: `onfido${SDK_ENV === 'auth' ? SDK_ENV : ''}.[name].min.js`,
```
ethan605, line 287: Ditto here (`'Auth'` rather than `'auth'`)

webpack.config.babel.js
```diff
@@ -320,10 +328,22 @@ const configDist = {
...
+                to: `${__dirname}/dist/core-sdk`,
+              },
+            ],
+          }),
+        ]
+      : []),
     new MiniCssExtractPlugin({
       filename: 'style.css',
-      chunkFilename: 'onfido.[name].css',
+      chunkFilename: `onfido${SDK_ENV === 'auth' ? SDK_ENV : ''}.[name].css`,
```
ethan605, line 346: Ditto here (`'Auth'` rather than `'auth'`)

webpack.config.babel.js
```diff
@@ -299,7 +307,7 @@ const configDist = {
               sourceMap: true,
               terserOptions: {
                 output: {
-                  preamble: `/* Onfido SDK ${packageJson.version} */`,
+                  preamble: `/* Onfido${SDK_ENV} SDK ${packageJson.version} */`,
```
ethan605, line 310: Ditto here (`'Auth'` rather than `'auth'`)

.gitignore
```diff
@@ -23,4 +24,4 @@ dist/previewer*
 dist/*~*
 test/mock-server/*.pem
 test/mock-server/frontend/
-coverage/
+coverage/
```
ethan605, line 27: Should be re-saved to append a new-line character here

package.json
```diff
@@ -10,15 +10,16 @@
...
     "build": "webpack --quiet --display-error-details --bail",
+    "build:auth": "SDK_ENV=auth webpack --quiet --display-error-details --bail",
     "build:dev": "NODE_ENV=development npm run build",
     "build:test": "NODE_ENV=test npm run build && npm run mock-server:build",
-    "prebuild": "npm run clean && npm run mkdirs && npm run license-checker && npm run check && npm run test",
     "postbuild": "dts-bundle-generator src/types/index.ts --out-file lib/index.d.ts",
     "mkdirs": "mkdir -p dist && mkdir -p lib",
     "clean": "rimraf lib dist",
-    "prepare": "npm run clean && npm run check && npm run build",
+    "prepare": "npm run clean && npm run build",
```
ethan605, line 22: Should be reverted to the original script

src/components/Complete/assets/success.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="56" height="56" viewBox="0 0 56 56" xmlns="http://www.w3.org/2000/svg">
+  <path d="M 28 52 C 34.625 52 40.625 49.313 44.969 44.969 C 49.313 40.625 52 34.625 52 28 C 52 24.688 51.328 21.531 50.113 18.66 C 48.898 15.789 47.141 13.203 44.969 11.031 C 42.797 8.859 40.211 7.102 37.34 5.887 C 34.469 4.672 31.313 4 28 4 C 21.375 4 15.375 6.688 11.031 11.031 C 6.688 15.375 4 21.375 4 28 C 4 34.625 6.688 40.625 11.031 44.969 C 15.375 49.313 21.375 52 28 52 C 28 52 28 52 28 52" fill="#3640f5"/>
+  <path d="M 24.94 35.61 L 38.27 22.28 L 36.38 20.39 L 24 32.78 L 18.28 27.06 L 16.39 28.95 L 23.06 35.62 C 23.31 35.86 23.65 36 24 36 C 24.35 36 24.69 35.86 24.94 35.61 Z" fill="#ffffff"/>
+</svg>
```
ethan605, line 4: Ditto here (re-save to have new-line character)

src/types/steps.ts
```diff
@@ -47,6 +49,8 @@ export type StepOptionWelcome = {
   nextButton?: string
 }
 
+export type StepOptionAuth = { string: any }
```
ethan605, line 52: This should be more specific, what kinds of `string` value should be?

src/components/Auth/AuthConfig.js
```diff
@@ -0,0 +1,173 @@
+import { loader } from './assets/loader'
+import { success } from './assets/success'
+import { borderRadius, color } from '@onfido/castor'
+
+export const Config = (function () {
```
ethan605, line 5: Is there any specific reason that this component wasn't written in TypeScript (while most of the others are)?

src/components/Auth/Auth.js
```diff
@@ -0,0 +1,95 @@
+import { h, Component } from 'preact'
+import { AuthCheckProcessor } from './AuthCheckProcessor'
+import { FaceTecSDK } from '../../../core-sdk/FaceTecSDK.js/FaceTecSDK'
+import { Config } from './AuthConfig'
+import { FaceTecStrings } from './assets/FaceTecStrings'
+import { Button } from '@onfido/castor-react'
+
+export default class AuthCapture extends Component {
```
ethan605, line 8: Ditto here (non-TypeScript component)

ethan605 (changes requested):
49/67 files reviewed. Sorry for being harsh here, let's follow the best practices to introduce stable & reliable new features :D

.prettierignore
stefaniacardenas: I think the name `core-sdk` could generate confusion given we already use the term "core SDK" on mobile SDKs to refer to one of the SDK modules

lib/index.d.ts
```diff
@@ -122,6 +128,21 @@ export declare type UICustomizationOptions = {
...
+	authButtonCornerRadius?: string;
+	authFrameColor?: string;
+	authButtonTextHighlightColor?: string;
+	authButtonTextDisabledColor?: string;
+	authButtonTextNormalColor?: string;
+	authButtonColorNormal?: string;
+	authButtonColorDisabled?: string;
+	authButtonColorPressed?: string;
+	authFeedbackBarColor?: string;
+	authFeedbackBarTextColor?: string;
```
stefaniacardenas, line 145: Has this been documented in the readme?

src/components/Auth/Auth.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+  componentDidUpdate(prevProps, prevState) {
+    if (
+      Object.keys(prevState.authConfig).length === 0 &&
+      Object.keys(this.state.authConfig).length > 0 &&
+      FaceTecSDK.getStatus() === 0
+    ) {
+      this.initFaceTec()
+    }
+  }
```
stefaniacardenas, line 22: I agree with @ethan605'c comment plus I would use hooks

src/components/Auth/Auth.js
```diff
@@ -0,0 +1,95 @@
...
+
+  initFaceTec = () => {
+    FaceTecSDK.setResourceDirectory('../../../core-sdk/FaceTecSDK.js/resources')
+    FaceTecSDK.setImagesDirectory('../../../core-sdk/FaceTec_images')
```
stefaniacardenas, line 26: You can create an alias in `webpack.config.babel.js` so you can reference core-sdk much more easily by doing something like `'~/core-sdk/FaceTecSDK.js/resources'`

src/components/Auth/AuthCheckProcessor.tsx
```diff
@@ -0,0 +1,170 @@
...
+  ) {
+    this.sdkToken = sdkToken
+    this.success = false
+    this.nextStep = nextStep
+    this.events = events
+
+    //
+    // Part 1:  Starting the FaceTec Session
+    //
+    const facetecsession = new FaceTecSDK.FaceTecSession(this, sessionToken)
```
stefaniacardenas, line 30: Why not using camelCase for this variable?

src/components/Auth/AuthCheckProcessor.tsx
```diff
@@ -0,0 +1,170 @@
...
+        sdk_metadata: {
+          system: {
+            fingerprint: '',
+            model: '',
+            manufacturer: '',
+            brand: '',
+            product: '',
+            hardware: '',
+          },
+        },
```
stefaniacardenas, line 73: Why are these empty strings?

src/components/Auth/AuthCheckProcessor.tsx
```diff
@@ -0,0 +1,170 @@
...
+          })
+          FaceTecSDK.unload(() => {})
+        } catch {
+          // CASE:  Parsing the response into JSON failed --> You define your own API contracts with yourself and may choose to do something different here based on the error.  Solid server-side code should ensure you don't get to this case.
+          this.success = null
+          const message =
+            'Exception while handling API response, cancelling out.'
+          this.events?.emit('error', { type: 'exception', message })
+          faceScanResultCallback.cancel()
+          this.nextStep()
```
stefaniacardenas, line 132: What's the `nextStep` when this request fails?

src/components/Auth/AuthCheckProcessor.tsx
```diff
@@ -0,0 +1,170 @@
...
+          FaceTecSDK.unload(() => {})
+        }
+      }
+    }
+
+    this.latestNetworkRequest.onerror = () => {
+      // CASE:  Network Request itself is erroring --> You define your own API contracts with yourself and may choose to do something different here based on the error.
+      const message = 'XHR error, cancelling.'
+      this.events?.emit('error', { type: 'exception', message })
+      console.log(message)
```
stefaniacardenas, line 142: Did you intend to leave this log here?

src/components/Auth/AuthCheckProcessor.tsx
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  //
+  // Part 10:  This function gets called after the FaceTec SDK is completely done.  There are no parameters because you have already been passed all data in the processSessionWhileFaceTecSDKWaits function and have already handled all of your own results.
+  //
+  onFaceTecSDKCompletelyDone = () => {
+    //
+    // DEVELOPER NOTE:  onFaceTecSDKCompletelyDone() is called after you signal the FaceTec SDK with success() or cancel().
+    // Calling a custom function on the Sample App Controller is done for demonstration purposes to show you that here is where you get control back from the FaceTec SDK.
+    console.log('complete!')
```
stefaniacardenas, line 168: This log is too generic. Remove or replace with something more meaningful

src/components/Auth/AuthConfig.js
```diff
@@ -0,0 +1,173 @@
...
+    this.currentCustomization = defaultCustomization
+
+    // Set Animated Customization
+    const loaderSVG = document.createElementNS(
+      'http://www.w3.org/2000/svg',
+      'svg'
+    )
+    loaderSVG.setAttribute('viewBox', '0 0 56 56')
+    loaderSVG.classList.add('experiment-svg')
+    loaderSVG.innerHTML = loader
```
stefaniacardenas, line 150: Using innerHTML in the client can be a security concern. Please find an alternative or make sure the html is properly sanitised.

src/components/Auth/AuthIntro.tsx
```diff
@@ -0,0 +1,44 @@
...
+          onClick={nextStep}
+          className={classNames(theme['button-centered'], theme['button-lg'])}
+        >
+          {translate('auth_intro.button_primary')}
+        </Button>
+      </div>
+    </div>
+  )
+}
+
```
stefaniacardenas, line 40: I would suggest using the `ScreenLayout` component to enforce a consistent layout

src/components/Auth/AuthIntro.tsx
```diff
@@ -0,0 +1,44 @@
...
+export default trackComponent(
+  //@ts-ignore
+  compose(localised, withCrossDeviceWhenNoCamera)(Intro)
+)
```
stefaniacardenas, line 44: Why not
```suggestion
export default compose(
  appendToTracking,
  localised,
  withCrossDeviceWhenNoCamera
)(Intro)
```

src/components/Auth/Lazy.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+const Loading = localised(({ translate }) => (
+  <div className={style.loading}>
+    {translate('generic.lazy_load_placeholder')}
+  </div>
+))
+
+const AsyncCrossDevice = asyncComponent(
+  //@ts-ignore
+  () => import(/* webpackChunkName: "vendor" */ './Auth.js'),
```
stefaniacardenas, line 14: I would rename this to something more specific
```suggestion
  () => import(/* webpackChunkName: "authVendor" */ './Auth.js'),
```

src/components/Auth/assets/loader.tsx
```diff
@@ -0,0 +1,2 @@
...
...7c.49-.29.95-.6 1.41-.92.32-.22.36-.67.1-.95-1.25-1.33-2.6-2.63-4.04-3.86-1.49-1.27-3.5-1.77-5.42-1.35-2.04.46-4.01 1.01-5.89 1.69-2.8 1-5.49 2.28-8.01 3.78-.39.23-.78.48-1.16.73l-.2.13c-.25.17-.25.55.02.71 6.88 4.17 15.79 4.47 23.19.04z" fill="#99a3f4"/><path class="right" d="M45.45 40.37c-.32-.46-.63-.92-.92-1.41-4.43-7.39-4.13-16.31.03-23.2.16-.27.54-.27.71-.02.04.07.09.13.13.2.25.38.5.77.73 1.16 1.5 2.52 2.79 5.21 3.78 8.01.68 1.88 1.24 3.85 1.69 5.89.43 1.92-.08 3.93-1.35 5.42-1.24 1.44-2.53 2.79-3.86 4.04-.27.27-.72.22-.94-.09z" fill="#99a3f4"/><path d="M28.087 19.58c-1.782-.003-3.519.56-4.961 1.607-1.441 1.046-2.514 2.524-3.063 4.219-.549 1.695-.546 3.521.008 5.214.554 1.693 1.632 3.167 3.077 4.209 1.445 1.042 3.184 1.599 4.965 1.591 1.782.003 3.519-.56 4.961-1.606 1.441-1.047 2.514-2.525 3.063-4.22.549-1.695.546-3.521-.008-5.214-.554-1.693-1.632-3.167-3.077-4.209-1.445-1.042-3.184-1.599-4.965-1.591zM15.73 45.45c.46-.32.92-.63 1.41-.92 7.4-4.43 16.31-4.13 23.2.03.27.16.27.54.02.71-.07.04-.13.09-.2.13-.38.25-.77.5-1.16.73-2.52 1.5-5.21 2.79-8.01 3.78-1.88.68-3.85 1.24-5.89 1.69-1.92.43-3.93-.08-5.42-1.35-1.44-1.24-2.79-2.53-4.04-3.86-.27-.27-.23-.72.09-.94z" fill="#3640f5"/>'
```
stefaniacardenas, line 2: What is this?

stefaniacardenas (changes requested):

.prettierignore
natasharsilva: I can update the name to `auth-sdk` :) if you think it's worth it, I'll also move the files into a Facetec folder and update this line to `auth-sdk/FaceTec`

tsconfig.json
natasharsilva: Apparently yes, this is due to the need to use dynamic imports in `StepComponentMap.ts`. Is this a deal-breaker? Is there a way around it?

webpack.config.babel.js
natasharsilva: not 100% if this is what Carla intended here, but I believe this URL will change to a new one, will check

carlagn: `OLD_AUTH` was referring to an old url we were using to test the earlier stages of this implementation- it went by without an update to the name, which is due 😓  my bad. This URL is for our own production backend for auth, so we should keep using it.

src/components/Auth/AuthConfig.js
carlagn: Previous use-case provided by FaceTec was using a js file, and transpiling it to TS was not working 😒  I've since fixed this and will upload the `.tsx` version of this file, with additional changes to cater to custom UI provided by the client.

src/components/Auth/AuthCheckProcessor.tsx
carlagn: nextStep should be whatever the integrator defines. We're calling `nextStep` here as well to keep consistency, but regarding behaviour it might be worth a revision.

src/components/Auth/AuthCheckProcessor.tsx
carlagn: Yes. We're logging all major events, which should include exceptions, session inits, etc. This `console.log` in specific is to keep consistent.

src/components/Auth/AuthConfig.js
carlagn: I've tried with multiple different ways to get the `SVGElement` to be accepted/rendered, but nothing except this works. This is also the method used by them on their examples, so I'm begrudgingly suspecting that it really is the only way... is there any recommendation you can provide to further sanitise this piece of code?

src/components/Auth/AuthIntro.tsx
carlagn: On doing this change, I'm getting the following TS error for `localised`:

```Type '"language"' is not assignable to type '"trackScreen"'.ts(2345)```

File: src/components/Auth/assets/loader.tsx (hunk 29 above)
Comment (carlagn) [reply src/components/Auth/assets/loader.tsx:2]: This is the custom svg loader. This is set as a const as the only way to fetch it/modify it inside `AuthConfig.tsx` is by setting it as a variable. 
This SVG is later subject to CSS animations to move according to the designs. It's used for "loading" use-cases.

File: src/components/Auth/AuthIntro.tsx (hunk 27 above)
Comment (ethan605) [reply src/components/Auth/AuthIntro.tsx:44]: Okay firstly, because you've consumed `useLocales` hook, you don't need `localised` HoC anymore
Secondly, I would suggest avoiding `compose` because it encourages HoCs overuse and hides props collisions. So in this situation you simply need:

```suggestion
export default trackComponent(withCrossDeviceWhenNoCamera(Intro))
```

and remove redundant imports

webpack.config.babel.js
carlagn: We used lowercase to keep all characters consistent. Mostly to avoid any mismatching if the SDK_ENV variable is set with all lower case rather than capitalised. I can change it if you'd prefer? although files generated in `dist` are named `onfidoAuth` (with capital A)

webpack.config.babel.js
kopijunkie: Maybe `LEGACY_AUTH` might be a better name for this?

kopijunkie (changes requested):
Additional question, since I couldn't add a comment to them, but are all the PNG files added here optimised already? 
FYI we've not implemented image optimisation & so far have had to manually optimise image files. I've only added SVG optimisation to our Github Actions workflow so far (didn't work when I tried via Webpack).

src/components/Auth/assets/animations.scss
```diff
@@ -0,0 +1,125 @@
+:global {
+  .experiment-svg {
```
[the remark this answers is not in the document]
carlagn: It needs to be in a global scope as the animations need to be available throughout the whole application. Since facetec renders its components on body root, we can't pinpoint a DOM area to set these css animations.

src/locales/de_DE/de_DE.json
```diff
@@ -1,4 +1,98 @@
 {
+    "auth_intro": {
+        "title": "Welcome back!",
+        "subtitle": "To make sure it’s you, please scan your face",
+        "line1": "You’ll be asked to keep your face within an oval frame.",
+        "line2": "Make sure your eyes are clearly visible.",
+        "button_primary": "Start camera"
+    },
```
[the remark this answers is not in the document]
carlagn: This object was not made available by lokalise :/ will ping rae for this.

webpack.config.babel.js
carlagn: For now I'll change it to `AUTH_URL` as this one is the production one we actually use, if you agree?

.eslintignore
carlagn: I've tried both methods to prettify the file but somehow I'm gettings this message even after running the file through prettier and format:
```
src/components/Auth/assets/animations.scss
 21:13  ✖  Unexpected whitespace after "("    function-parentheses-space-inside
 23:4   ✖  Unexpected whitespace before ")"   function-parentheses-space-inside
 (...)
 73:13  ✖  Unexpected whitespace after "("    function-parentheses-space-inside
 75:4   ✖  Unexpected whitespace before ")"   function-parentheses-space-inside
 ```
 I'm not sure why. when I remove the whitespaces the error transforms into:
 ```
  64:3   ✖  Expected indentation of 4 spaces   indentation                      
```
Could it be because I'm changing svg path info on these lines? they refer to:
```
    d: path(
      'M 24.94 35.61 L 38.27 22.28 L 36.38 20.39 L 24 32.78 L 18.28 27.06 L 16.39 28.95 L 23.06 35.62 C 23.31 35.86 23.65 36 24 36 C 24.35 36 24.69 35.86 24.94 35.61 Z'
    );
```

src/components/Auth/AuthCheckProcessor.tsx
```diff
@@ -0,0 +1,170 @@
...
+            product: '',
+            hardware: '',
+          },
+        },
+      },
+    }
+
+    //
+    // Part 5:  Make the Networking Call to Your Servers.  Below is just example code, you are free to customize based on how your own API works.
+    //
```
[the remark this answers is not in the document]
kopijunkie: Yeah. Was wondering about it since there's a lot of information & example code in here, which is not a bad thing.

src/locales/de_DE/de_DE.json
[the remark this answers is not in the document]
kopijunkie: @raemarielee FYI ^

src/components/Auth/Auth.js
carlagn: Please refer to [this PR](https://github.com/onfido/onfido-sdk-ui/pull/1431/commits/44cae6abcdb83da897695e7e89ef1912be55fdca) to take a look at this implementation :) cheers ⭐

src/locales/de_DE/de_DE.json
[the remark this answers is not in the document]
carlagn: @kopijunkie I've spoken with rae regarding these- overall they will **not** be added to the SDK itself, as the `AuthIntro` screen is meant to be implemented by the integrator themselves (or, in our our case, most likely a screen on Client Trial App). I've since removed the `AuthIntro` screen from the `Auth` flow on StepComponent, and will proceed to erase these entries to keep it tidy.

.eslintignore
carlagn: @ethan605 made a go at the `animations.scss` file on [this PR](https://github.com/onfido/onfido-sdk-ui/pull/1431/commits/481fa8176ea881eacfc0410c5d7b677ca6837ac8) 😁

.gitignore
carlagn: I've updated this file a couple of times locally and pushed it with the additional new-line at the end, but I think gitlab is removing it? [(example)](https://github.com/onfido/onfido-sdk-ui/pull/1414/commits/3abecc112a2491f3cd434d9e26dffd04ca43cfb8) Am I missing something?

tsconfig.json
carlagn: @ethan605 I've updated the CHANGELOG with this info **only**. Will all additional information regarding this module in specific be added on this file as well? i.e. separate bundle builds, auth flow explanation and screens added, etc.

src/types/steps.ts
ethan605: Should be, just not sure about the granularity. Any ideas @kopijunkie @stefaniacardenas :thinking:

src/demo/demoUtils.ts
```diff
@@ -145,7 +145,11 @@ export const getInitSdkOptions = (): SdkOptions => {
       ? SAMPLE_LOCALE
       : queryParamToValueString.language
 
-  const steps: Array<StepConfig> = [{ type: 'welcome' }]
+  const steps: Array<StepConfig> = [
+    { type: 'welcome' },
+    { type: 'userConsent' },
+    { type: 'auth' },
```
stefaniacardenas, line 151: This step should only be requested if it's we are testing auth, otherwise there will be a log in the console complaining about the fact that auth is not a step.

src/locales/es_ES/es_ES.json
```diff
@@ -1,4 +1,91 @@
...
+            "move_back": "Ahora muévase hacia atrás",
+            "move_close": "Ahora acérquese",
+            "move_closer": "Acérquese",
+            "not_looking_straight": "Mire hacia adelante",
+            "steady": "No se mueva"
+        }
+    },
+    "auth_error": {
+        "cam_encryption": {
+            "body": "Esta aplicación bloquea las configuraciones sospechosas de las cámaras web. <a href='https:\/\/livenesscheckhelp.com\/' target='_blank' style='text-decoration:underline;'>Más información<\/a>.",
```
stefaniacardenas, line 44: This could leave us open to XSS attacks. You need to use an XML tag here and the `parseTags` utils function. Take a look at the CountrySelector component to get an idea of how it's done or message me or @kopijunkie. It needs to be changed also on Lokalise

stefaniacardenas (changes requested):
I've requested some changes in the comment. The most important ones are the ones related to the use of `innerHTML` and the use of HTML tags in the locale strings. Below, are some additional questions that I could not add to the actual file they belonged to.

1) Can we delete the following files?
  * auth-sdk/FaceTec/FaceTecSDK.js/ThirdPartyLicenses.txt
  * auth-sdk/FaceTec/FaceTec_images/FaceTec_branding_logo_id_check.png
2) Do we need all those png files from Facetech?
3) Can we include the Facetec code in a public repo/ front end facing library?

stefaniacardenas:
Two more points:
* Before merging this we should do a test release candidate to find out how long it takes to download `onfido-sdk-ui` with this change. If it takes too long we should probably publish the auth bundle as a separate NPM module.
* I can see that this feature has not been tested. I would suggest adding at least some UI test before releasing such a big feature

src/components/Auth/AuthCheckProcessor.ts
```diff
@@ -0,0 +1,173 @@
...
+            product: '',
+            hardware: '',
+          },
+        },
+      },
+    }
+
+    //
+    // Part 5:  Make the Networking Call to Your Servers.  Below is just example code, you are free to customize based on how your own API works.
+    //
```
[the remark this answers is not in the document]
carlagn: As mentioned here to kimmi: https://github.com/onfido/onfido-sdk-ui/pull/1414#discussion_r629281729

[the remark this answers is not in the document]
carlagn: I'll re-edit the comments to better address whoever would benefit from these comments. Or would you rather I remove altogether?

src/components/Auth/AuthConfig.tsx
```diff
@@ -0,0 +1,174 @@
...
+      uiDef.authAccentColor
+
+    // Set Animated Customization
+    const loaderSVG = document.createElementNS(
+      'http://www.w3.org/2000/svg',
+      'svg'
+    )
+    loaderSVG.setAttribute('viewBox', '0 0 56 56')
+    loaderSVG.classList.add('experiment-svg')
+    loaderSVG.innerHTML = loader
```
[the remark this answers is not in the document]
carlagn: I did a `loaderSVG.innerHTML = sanitize(loader)` and unfortunately it doesn't show anything where the loader SVG was supposed to be :(

src/locales/es_ES/es_ES.json
carlagn: Are there any specific keys I should add the `parseTags` to? Or all of them?

src/components/Auth/AuthConfig.tsx
```diff
@@ -0,0 +1,174 @@
...
+        : color('background-disabled'),
+      authButtonColorPressed: dimMode
+        ? color('primary-200')
+        : color('background-action-active'),
+      authFeedbackBarColor: dimMode
+        ? color('neutral-900')
+        : color('background-overlay'),
+      authFeedbackBarTextColor: color('neutral-white'),
+    }
+    const uiDef = { ...uiDefaults, ...customUI }
```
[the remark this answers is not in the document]
carlagn: Yes. More explicit?

carlagn:
> I've requested some changes in the comment. The most important ones are the ones related to the use of `innerHTML` and the use of HTML tags in the locale strings. Below, are some additional questions that I could not add to the actual file they belonged to.
> 
> 1. Can we delete the following files?
> 
> * auth-sdk/FaceTec/FaceTecSDK.js/ThirdPartyLicenses.txt
> * auth-sdk/FaceTec/FaceTec_images/FaceTec_branding_logo_id_check.png
Yes. Updated this.
> 1. Do we need all those png files from Facetech?
I've left only essencial pngs for now, as we only need those related to liveness.
> 2. Can we include the Facetec code in a public repo/ front end facing library?
Short answer: yes. Long answer: https://onfido.slack.com/archives/C01M05YNGUU/p1620219320010200

src/components/Auth/AuthCheckProcessor.ts
```diff
@@ -0,0 +1,173 @@
...
+            console.error('Unexpected API response, cancelling out.')
+            faceScanResultCallback.cancel()
+            this.nextStep()
+          }
+
+          this.events?.emit('complete', {
+            type: 'complete',
+            ...responseObj,
+          })
+          FaceTecSDK.unload(() => {})
```
kopijunkie, line 125: Is there any chance that this `unload` method might fail? In which case would be good to log/catch the error if it happens.

src/types/steps.ts
```diff
@@ -47,6 +49,8 @@ export type StepOptionWelcome = {
   nextButton?: string
 }
 
+export type StepOptionAuth = { retries?: number }
```
kopijunkie, line 52: Is `retries` here referring to number of capture retries for the Auth step?

webpack.config.babel.js
```diff
@@ -122,6 +127,7 @@ const PROD_CONFIG = {
   SMS_DELIVERY_URL: 'https://telephony.onfido.com',
   PUBLIC_PATH: `https://assets.onfido.com/web-sdk-releases/${packageJson.version}/`,
   USER_CONSENT_URL: 'https://assets.onfido.com/consent/user_consent.html',
+  AUTH_URL: 'https://edge.api.onfido.com/v3',
```
kopijunkie, line 130: Is this only available on Prod environment? How about Staging/Dev environment (see `STAGING_CONFIG` [below](https://github.com/onfido/onfido-sdk-ui/blob/f9aaf6e980eddd7130e7e96ffe32af4007747d31/webpack.config.babel.js#L152)) and the `TEST_E2E_CONFIG` for automated tests?

src/locales/es_ES/es_ES.json
kopijunkie: This would also cause any UI styling/customisation to not get applied since there isn't any class name for this link element.

kopijunkie (changes requested):
Some questions & changes to be made. 

I've also noticed that there seems to be an empty file in this PR's diffs at [`auth-sdk/FaceTec/FaceTecSDK.js/ThirdPartyLicenses.txt`](https://github.com/onfido/onfido-sdk-ui/blob/f9aaf6e980eddd7130e7e96ffe32af4007747d31/auth-sdk/FaceTec/FaceTecSDK.js/ThirdPartyLicenses.txt) - any reason that this empty file is needed?

src/locales/es_ES/es_ES.json
carlagn: I'm confused. When you say link element what are you referring to? And where would a class name be needed for this case?

kopijunkie: This part `<a href='https:\/\/livenesscheckhelp.com\/' target='_blank' style='text-decoration:underline;'>`. It's usually handled by `parseTags` & needs a `<fallback>` tag configured on Lokalise side also. Let's pair on this. Will be easier than trying to explain in words. 😄

README.md
```diff
@@ -543,6 +543,30 @@ A number of options are available to allow you to customize the SDK:
...
+  | `authFrameColor`            | <ul><li>Change the color of the border and background surrounding the oval outline Auth session</li><li>Change background color of the Auth Overlay.</li><li>Change background color for the New User Guidance and Retry Screens.</li><li> Change background color of Result screen and Result animation foreground</li><li>Change background color of initial loading animation</li></ul> |
+  | `authButtonCornerRadius`    | <ul><li>Change corner radius of the action button's border during the New User Guidance and Retry Screens.</li><li>Change corner radius of the Auth Feedback Bar.</li></ul>                                                                                                                                                                                                                |
```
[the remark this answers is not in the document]
carlagn: After doing some digging on how to do unordered lists inside table cells, most suggestions bring up mixing markup with html for this use case in specific.

Using markup inside the table cell breaks the table due to the newlines.

src/components/Auth/AuthCheckProcessor.ts
```diff
@@ -0,0 +1,173 @@
...
+    //
+    // Part 4:  Get essential data off the FaceTecSessionResult
+    //
+    const parameters = {
+      face_scan: sessionResult.faceScan,
+      audit_trail_image: sessionResult.auditTrail[0],
+      low_quality_audit_trail_image: sessionResult.lowQualityAuditTrail[0],
+      metadata: {
+        sdk_source: 'onfido_web_sdk',
+        sdk_version: '1.0',
```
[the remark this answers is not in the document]
carlagn: For now it is a dummy. This in specific refers to the device info.

src/components/Auth/AuthCheckProcessor.ts
carlagn: This unload method needs a bit more digging since it's not doing what we assumed it was meant for. We have gone in contact with facetec to further investigate this. Irregardless of this, the unload function call or lack thereof should not be an issue for this stage on auth.

src/types/steps.ts
carlagn: Yes.

src/components/Auth/AuthCheckProcessor.ts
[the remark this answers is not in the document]
stefaniacardenas: I think it makes sense to leave the comments when they are reflecting what the code does. So here I would only have 
`Part 5:  Make the Networking Call to the Onfido Servers.` 

> Below is just example code

This should be removed because the code is not an example and if I change anything in that code it would break the integration (unless I know what I'm doing) :)

src/components/Auth/AuthCheckProcessor.ts
```diff
@@ -0,0 +1,173 @@
...
+    )
+    this.latestNetworkRequest.setRequestHeader(
+      'Content-Type',
+      'application/json'
+    )
+
+    this.latestNetworkRequest.onreadystatechange = () => {
+      //
+      // Part 6:  In our Sample, we evaluate a boolean response and treat true as success, false as "User Needs to Retry",
+      // and handle all other non-nominal responses by cancelling out.  You may have different paradigms in your own API and are free to customize based on these.
```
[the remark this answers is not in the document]
stefaniacardenas: > In our Sample
It's unclear which sample you (or they) are referring to here. My memory is so bad that in 3 months time I might be thinking by sample you mean [our sample app](https://github.com/onfido/onfido-sdk-web-sample-app/) 😅

src/components/Auth/AuthConfig.tsx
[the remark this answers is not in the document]
stefaniacardenas: Is `loader` a string or a node?

src/locales/es_ES/es_ES.json
stefaniacardenas: Btw, you can call it whatever you want not necessarily `<fallback>`. You can call it <cameraHelpLink> or whatever makes sense to you in this scenario

src/components/Auth/AuthCheckProcessor.ts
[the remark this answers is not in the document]
stefaniacardenas: Why do you need this dummy when you can use the actual SDK version?

src/components/Auth/assets/animations.scss
```diff
@@ -0,0 +1,133 @@
+/* stylelint-disable function-parentheses-space-inside */
+:global {
+  .loaderSvg {
```
stefaniacardenas, line 3: Nice. `onfidoLoaderSvg` would make it even more specific, so we don't end up with conflicts with the host app

src/components/Auth/assets/loader.tsx
stefaniacardenas: If `dompurify` doesn't work with this then we should bring this up with security because I'm pretty sure this is high risk.

stefaniacardenas (changes requested):

src/components/Auth/AuthConfig.tsx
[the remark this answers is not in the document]
stefaniacardenas: For some reason my previous comment disappeared, so pasting it here again
If `dompurify` doesn't work with this then we should bring this up with security because I'm pretty sure this is high risk.

src/components/Auth/AuthCheckProcessor.ts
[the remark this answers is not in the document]
carlagn: Absolutely. Will update these comments, thank you!

src/components/Auth/AuthConfig.tsx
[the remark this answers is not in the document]
carlagn: it's a string! I tried printing the "sanitized" value and whereas `loader = '<path....'`, `sanitize(loader) = ''`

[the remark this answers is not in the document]
stefaniacardenas: Security team gave us the green light on this so I will resolve this conversation

webpack.config.babel.js
carlagn: e2e tests: https://gitlab.eu-west-1.mgmt.onfido.xyz/onfido/biometrics/auto/facetec_tests/

src/components/Complete/style.scss
```diff
@@ -7,6 +7,6 @@
 .icon {
   height: 64 * $unit;
   background-size: contain;
-  background-image: url('./assets/complete.svg');
+  background-image: url('./assets/success.svg');
```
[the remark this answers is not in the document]
carlagn: I changed it to have the same look as auth at the time but this change shouldn't here, no. Updating this.

src/demo/demoUtils.ts
```diff
@@ -146,7 +147,14 @@ export const getInitSdkOptions = (): SdkOptions => {
       ? SAMPLE_LOCALE
       : queryParamToValueString.language
 
-  const steps: Array<StepConfig> = [{ type: 'welcome' }]
+  const steps: Array<StepConfig> = [
+    { type: 'welcome' },
+    { type: 'userConsent' },
```
stefaniacardenas, line 152: Tests are failing because `{ type: 'userConsent' }` has been introduced here

stefaniacardenas (changes requested):

src/components/Auth/AuthCheckProcessor.ts
carlagn: I'm still waiting for a reply back on this from FaceTec. I'd say it's not a blocker for this initial release, but will raise it later:)

stefaniacardenas (approved):

src/components/Auth/AuthCheckProcessor.ts
carlagn: @kopijunkie I have discussed this unload issue with Lee from FaceTec and it appears that we do not need this function call at all. This function should only be used to reset the resource directories setup, and nothing directly related to the session flow itself. I've removed these, and addressed the issue we had with re-beginning a new session, so the goal we were trying to achieve with this call has been resolved:)

kopijunkie (changes requested):
1 concern about some of the React Hook warnings & a couple of minor points. But otherwise looking good.

src/components/Auth/AuthConfig.tsx
```diff
@@ -0,0 +1,179 @@
...
+      authRetryScreenOvalColor: color('primary-500'),
+
+      authOvalColor: '#FFFFFF',
+      authTextColor: dimMode ? color('neutral-200') : color('content-main'),
+      authButtonCornerRadius: borderRadius('medium'),
+      //Dimming mode overrides the frameColor to #000000
+      authFrameColor: '#FFFFFF',
+      //Button colors for night/dim mode + day
+      authButtonTextHighlightColor: dimMode
+        ? color('primary-500')
```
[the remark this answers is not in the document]
carlagn: `content-inverse-on-action` is not being recognised on my side for a `Palette` type :/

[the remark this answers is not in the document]
kopijunkie: Will just mark as resolved. Was a minor nitpick & I recall there were some tokens on the Figma that didn't seem to be in Castor also when doing the UI customisation work. 🤔

kopijunkie (approved):
LGTM

src/components/Auth/assets/FaceTecStrings.js
```diff
@@ -0,0 +1,124 @@
+const feedback = (el) => `auth_capture.feedback.${el}`
+const presessionFeedback = (el) => `auth_capture_start.feedback.${el}`
+export const FaceTecStrings = (t) => ({
```
ethan605, line 3: Firstly, using a single letter `t` is quite an anti-pattern in JS, because it decreases the readability.
Secondly, you're assuming the passed translate function is safe and ready for your calls, but in real life, sometimes it isn't.

```suggestion
export const FaceTecStrings = (translate) => {
  // `translate` function checks
  if (!translate || typeof translate !== 'string') {
    // ...
  }
  
  return {
    FaceTec_accessibility_cancel_button: translate('auth_accessibility.back_button'),
    // ...
  }
}
```

ethan605: Also, please convert this into TypeScript :pray:

src/components/Router/StepComponentMap.ts
```diff
@@ -109,6 +123,10 @@ const buildCaptureStepComponents = (
     welcome: [Welcome],
     userConsent: [UserConsent],
     face: buildFaceComponents(faceStep, deviceHasCameraSupport, mobileFlow),
+    //@ts-ignore
```
ethan605, line 126: You don't need `@ts-ignore` here. Your `LazyAuth` component is TypeScript-friendly enough. All the other `@ts-ignore`'s in this file are just a workaround while waiting for the related components to be ported to TS.
```suggestion
```

ethan605 (changes requested):
Generally LGTM. Just need some minor changes

README.md
```diff
@@ -866,6 +919,17 @@ The custom options are:
   When disabled, it will forward the user to the cross-device flow in order to attempt to capture a video in another device. If the user is already in a mobile device and it does not support
   MediaRecorder, the unsupported browser error will be shown.
 
+#### auth
+
+This is Authentication step. If you have followed the guidelines specific to including auth, you'll have this step made available. In here, a loading screen is presented to the user to fetch all necessary resources to perform authentication.
```
superdevx64, line 926: Would a user never not have followed the guidelines specific to including auth? 

In other words: won't users always have this step available (or shouldn't they)?

README.md
```diff
@@ -866,6 +919,17 @@ The custom options are:
...
+
+After all resources are loaded, the session is initialised, and the authentication check begins. An oval frame of the camera will be present (if camera permissions are provided) and actionable elements will render, asking the user to place their face in frame, followed up by a different set of instructions for them to follow to successfully authenticate the user.
+
+If the user is not a match, or conditions are not good to successfuly authenticate, they will be asked to retry authentication. If authentication is not possible, the page will rollback to the previous step.
```
superdevx64, line 930: What implies "authentication is not possible"? Is it a certain number of unsuccessful attempts?

README.md
```diff
@@ -866,6 +919,17 @@ The custom options are:
...
+
+**The Authentication module is currently a BETA feature.**
```
superdevx64, line 932: I would place this higher up, and make a note with a ⚠️ emoji, consistently with other similar notices

README.md
```diff
@@ -541,6 +570,30 @@ The Web SDK supports customization options across the SDK screen including text,
...
+  | ------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+  | `authTextColor`                | <ul><li>Change color of text displayed on the New User Guidance and Retry Screens (not including the action button text).</li><li>Change color of text displayed on the Result Screen.</li><li>Change color of loading spinner foreground fill and message text shown to the user while the camera loads.</li></ul> |
+  | `authButtonTextHighlightColor` | Change color of the action button's text when the button is pressed during the New User Guidance and Retry Screens.                                                                                                                                                                                                 |
```
superdevx64, line 594:
```suggestion
  | `authButtonTextHighlightColor` | Change color of the action button's text when the button is pressed during the New User Guidance and Retry screens.                                                                                                                                                                                                 |
```

README.md
```diff
@@ -541,6 +570,30 @@ The Web SDK supports customization options across the SDK screen including text,
...
+  | `authButtonTextDisabledColor`  | Change color of the action button's text when the button is disabled during the New User Guidance and Retry Screens.                                                                                                                                                                                                |
```
superdevx64, line 595:
```suggestion
  | `authButtonTextDisabledColor`  | Change color of the action button's text when the button is disabled during the New User Guidance and Retry screens.                                                                                                                                                                                                |
```

README.md
```diff
@@ -541,6 +570,30 @@ The Web SDK supports customization options across the SDK screen including text,
...
+  | `authButtonTextNormalColor`    | Change color of the action button's text during the New User Guidance and Retry Screens                                                                                                                                                                                                                             |
```
superdevx64, line 596:
```suggestion
  | `authButtonTextNormalColor`    | Change color of the action button's text during the New User Guidance and Retry screens                                                                                                                                                                                                                             |
```

README.md
```diff
@@ -123,6 +123,12 @@ You can include the library as a regular script tag on your page:
 <script src="dist/onfido.min.js"></script>
 ```
 
+⚠️ Note: The above import does **not** include the Auth module. To include it, use:
+
+```html
+<script src="dist/onfidoAuth.min.js"></script>
```
superdevx64, line 129: Would it be worth moving this further up and just having it as part of the same code snippet as `<script src="dist/onfido.min.js"></script>`?

README.md
```diff
@@ -146,6 +152,15 @@ import { init } from 'onfido-sdk-ui'
 // commonjs style require
 var Onfido = require('onfido-sdk-ui')
 ```
+⚠️ Note: The above import does **not** include the Auth module. To include it, use:
```
superdevx64, line 156: Similar comment to above

README.md
```diff
@@ -227,6 +242,18 @@ For the face step an object is returned with the `variant` used for the face cap
     }
 ```
 
+For the Auth step a data object is returned with parameters `sucess`, `token`, `type`, and `uuid`. The `success` variable informs whether or not the user was authenticated successfuly, whereas `token` is a JWT that can be used to validate the user authentication.
```
superdevx64, line 246: Is it a JSON object? I'd say that rather than "data"

superdevx64, line 246:
```suggestion
For the Auth step a data object is returned with parameters `success`, `token`, `type`, and `uuid`. The `success` variable informs whether or not the user was authenticated successfuly, and `token` is a JWT that can be used to validate the user authentication.
```

superdevx64, line 246: Consider linking to literature about JWTs

README.md
```diff
@@ -227,6 +242,18 @@ For the face step an object is returned with the `variant` used for the face cap
...
+
+**Example of an auth `onComplete` data callback:**
+```javascript
+    {
+      "success": true,
+      "token": "eyJhbGciOiJSUz...",
+      "type": "complete",
+      "uuid": "b3b9142d-3071-401d-821b-17ab134d4798"
```
superdevx64, line 255: Sanity question: is this a fake UUID?

README.md
```diff
@@ -237,6 +264,8 @@ Callback that fires when an error occurs. The callback returns the following err
   - timeout and server errors
   - authorization
   - invalid token
+  - [Auth] exception handling API response
+  - [Auth] exception handling API response
```
superdevx64, line 270: Duplicated line?

README.md
```diff
@@ -541,6 +570,30 @@ The Web SDK supports customization options across the SDK screen including text,
...
+  | `authAccentColor`           | <ul><li>Change color of activity indicator animation shown during server-side work</li><li> Change background color of result animation</li><li> Change background color of the upload progress bar</li> </ul>                                                                                                                                                                             |
+  | `authFrameColor`            | <ul><li>Change the color of the border and background surrounding the oval outline Auth session</li><li>Change background color of the Auth Overlay.</li><li>Change background color for the New User Guidance and Retry Screens.</li><li> Change background color of Result screen and Result animation foreground</li><li>Change background color of initial loading animation</li></ul> |
```
superdevx64, line 578:
```suggestion
  | `authFrameColor`            | <ul><li>Change the color of the border and background surrounding the oval outline Auth session</li><li>Change background color of the Auth overlay.</li><li>Change background color for the New User Guidance and Retry screens.</li><li> Change background color of Result screen and Result animation foreground</li><li>Change background color of initial loading animation</li></ul> |
```

README.md
```diff
@@ -541,6 +570,30 @@ The Web SDK supports customization options across the SDK screen including text,
...
+  | `authButtonCornerRadius`    | <ul><li>Change corner radius of the action button's border during the New User Guidance and Retry Screens.</li><li>Change corner radius of the Auth Feedback Bar.</li></ul>                                                                                                                                                                                                                |
```
superdevx64, line 579:
```suggestion
  | `authButtonCornerRadius`    | <ul><li>Change corner radius of the action button's border during the New User Guidance and Retry screens.</li><li>Change corner radius of the Auth Feedback Bar.</li></ul>                                                                                                                                                                                                                |
```

README.md
```diff
@@ -541,6 +570,30 @@ The Web SDK supports customization options across the SDK screen including text,
...
+
+  | Auth Module Colors (unique) | Description                                                                                                              |
+  | --------------------------- | ------------------------------------------------------------------------------------------------------------------------ |
+  | `authDualSpinnercolor`      | Change animated Auth Progress Spinner strokes.                                                                           |
+  | `authRetryScreenovalcolor`  | Change oval's stroke that overlays the Ideal image example during the first Retry Screen.                                |
```
superdevx64, line 584:
```suggestion
  | `authRetryScreenovalcolor`  | Change oval's stroke that overlays the ideal image example during the first Retry screen.                                |
```

README.md
```diff
@@ -541,6 +570,30 @@ The Web SDK supports customization options across the SDK screen including text,
...
+  | `authOvalColor`             | Change color of Auth Oval outline                                                                                        |
+  | `authButtonColorNormal`     | Change background color of the action button during the New User Guidance and Retry Screens.                             |
```
superdevx64, line 586:
```suggestion
  | `authButtonColorNormal`     | Change background color of the action button during the New User Guidance and Retry screens.                             |
```

README.md
```diff
@@ -541,6 +570,30 @@ The Web SDK supports customization options across the SDK screen including text,
...
+  | `authButtonColorDisabled`   | Change background color of the action button when the button is disabled during the New User Guidance and Retry Screens. |
```
superdevx64, line 587:
```suggestion
  | `authButtonColorDisabled`   | Change background color of the action button when the button is disabled during the New User Guidance and Retry screens. |
```

README.md
```diff
@@ -541,6 +570,30 @@ The Web SDK supports customization options across the SDK screen including text,
...
+  | `authButtonColorPressed`    | Change background color of the action button when the button is pressed during the New User Guidance and Retry Screens.  |
```
superdevx64, line 588:
```suggestion
  | `authButtonColorPressed`    | Change background color of the action button when the button is pressed during the New User Guidance and Retry screens.  |
```

README.md
```diff
@@ -541,6 +570,30 @@ The Web SDK supports customization options across the SDK screen including text,
...
+  | `authButtonTextDisabledColor`  | Change color of the action button's text when the button is disabled during the New User Guidance and Retry Screens.                                                                                                                                                                                                |
+  | `authButtonTextNormalColor`    | Change color of the action button's text during the New User Guidance and Retry Screens                                                                                                                                                                                                                             |
+  | `authFeedbackBarTextColor`     | Change color of text displayed within the Auth Browser SDK Feedback Bar                                                                                                                                                                                                                                             |
```
superdevx64, line 597:
```suggestion
  | `authFeedbackBarTextColor`     | Change color of text displayed within the Auth Browser SDK Feedback bar                                                                                                                                                                                                                                             |
```

README.md
superdevx64, line 926:
```suggestion
This is the authentication step. If you have followed the guidelines specific to including authentication, you'll have this step made available. In here, a loading screen is presented to the user to fetch all necessary resources to perform authentication.
```

src/components/Auth/Auth.tsx
```diff
@@ -0,0 +1,111 @@
...
+          })
+        }
+      }
+      const body = {
+        sdk_type: 'onfido_web_sdk',
+      }
+      XHR.send(JSON.stringify(body))
+    }
+    if (FaceTecSDK.getStatus() === 1) setSessionInit(true)
+    if (FaceTecSDK.getStatus() === 0 && !sessionInit) {
```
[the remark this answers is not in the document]
carlagn: No. Once the user goes through the auth session once, the sdk does not need to be initialised again. I just made a point to set session state according to the value of the sdk status. The hook for `authConfig.token` will handle authentication session inits.

README.md
carlagn: No. This step is only available for clients that do require this. We have different imports for IDV and IDV w/ Auth, so the integrator that does _not_ use auth, won't be hindered by the npm size increase.

README.md
carlagn: I placed here following on poa's beta feature warning. Should I put it right at the first place it's mentioned?

src/components/Auth/Auth.tsx
```diff
@@ -0,0 +1,111 @@
...
+            ...response,
+            production_key_text: JSON.parse(atob(response.production_key_text)),
+          })
+        }
+      }
+      const body = {
+        sdk_type: 'onfido_web_sdk',
+      }
+      XHR.send(JSON.stringify(body))
+    }
```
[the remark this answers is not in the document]
carlagn: I refactored according to your suggestion and added the new function to `http.ts` file on `/utils`. I wasn't 100% sure where I could add that made sense, or if I should have a standalone file for it. If you could take a look when you can at the file I'd greatly appreciate it 🙇‍♀️

src/components/Auth/Auth.tsx
```diff
@@ -0,0 +1,111 @@
...
+      )
+      if (authConfig.token && !sessionInit) initFaceTec()
+      else getConfig()
+    } else if (authConfig.token.length === 0) {
+      getConfig()
+    }
+  }, [sessionInit, props, authConfig, props.translate, onLivenessCheckPressed])
+
+  useEffect(() => {
+    if (authConfig.token && sessionInit) onLivenessCheckPressed()
```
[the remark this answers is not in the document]
ethan605: As I can see this wasn't updated

src/components/Auth/AuthConfig.tsx
```diff
@@ -0,0 +1,179 @@
...
+import { FaceTecSDK } from '~auth-sdk/FaceTecSDK.js/FaceTecSDK'
+import { UICustomizationOptions } from '../../types/ui-customisation-options'
+
+const ConfigBuilder = () => {
+  const BaseURL = process.env.AUTH_URL
+  return {
+    BaseURL,
+    getAuthCustomization,
+  }
+}
```
ethan605, line 14: Sorry for not spotting this earlier, but IMO simple exported const could serve you well here, instead of wrapping them in an additional object like this.

src/components/Auth/AuthConfig.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+const getAuthCustomization = (
+  dimMode: boolean,
+  customUI?: UICustomizationOptions
+) => {
+  const uiDefaults = {
```
ethan605, line 20: Minor optimisation tweak here: if `uiDefaults` won't be dynamically manipulated during run-time, you might want to declare it outside of this method.

src/components/Auth/assets/FaceTecStrings.tsx
```diff
@@ -0,0 +1,177 @@
...
+  steady_count_1: 'auth_capture_start.feedback.steady_count_1',
+  steady_count_2: 'auth_capture_start.feedback.steady_count_2',
+  steady_count_3: 'auth_capture_start.feedback.steady_count_3',
+  remove_sunglasses: 'auth_capture_start.feedback.remove_sunglasses',
+  neutral_expression: 'auth_capture_start.feedback.neutral_expression',
+  conditions_too_bright: 'auth_capture_start.feedback.conditions_too_bright',
+  conditions_too_dark: 'auth_capture_start.feedback.conditions_too_dark',
+}
+
+export const FaceTecStrings = (translate: TranslateCallback) => ({
```
ethan605, line 30: Sorry I'm lazy here, but the best way should be to fully declare the return type for better consumption.

```suggestion
export const FaceTecStrings = (translate: TranslateCallback): Record<string, string> => ({
```

ethan605: As I can see, still no `translate` validations here

ethan605 (changes requested):

README.md
superdevx64: Yes I would do

src/components/Auth/AuthConfig.tsx
carlagn: uiDefaults is dependent on the `dimMode` boolean to set the values, so we should leave it here, I think?

src/components/Auth/assets/FaceTecStrings.tsx
carlagn: Sorry, I think I'm confused. `translate` validations? referring to https://github.com/onfido/onfido-sdk-ui/blob/b17e9f2aff972d422c61b13496647700abf57600/src/components/utils/localesMapping.ts#L3-L16 ?

README.md
carlagn: Added it at the `4.1 HTML Script Tag Include` where auth is mentioned for the first time in the README

src/components/Auth/assets/FaceTecStrings.tsx
ethan605: It was a suggestion before you converted this into TypeScript, so I think it's valid now.

README.md
superdevx64: OK, I think we say that in one of the config setting descriptions but it may be worth reiterating it here (i.e. describe what it means to be not possible, what the default is, what the user can do to change that)

ethan605 (approved):
LGTM! Good job @carlagn :tada:

ethan605:
Ah one more thing @carlagn, re:

> 29 out of the 68 file changes reported originate from the core-sdk facetec sdk folder. This folder is the core functionality of the Authentication provided from FaceTec's technology. This folder includes wasm, js, and ts files for the development part, plus image and binary resources for the rendering of the screens.

Did you document somewhere "How to upgrade FaceTec SDK to a newer version"? Just ignore me if you already did it.

carlagn:
> Did you document somewhere "How to upgrade FaceTec SDK to a newer version"? Just ignore me if you already did it.

I didn't. It's not something we've tackled yet, but if a need to upgrade the FaceTec SDK comes up, we'll upgrade it accordingly. I will document it for internal purposes.

PR #1373, opened 2021-03-15
Improvement/CX-6310 - Find step util

ethan605:
# Problem

Currently, for a given `steps: StepConfig[]`, we're using a _hard_ way to retrieve a specific step:

```typescript
  const documentStep = steps.find(
    (step) => step.type === 'document'
  ) as StepConfigDocument
```

which is error-prone and somehow ugly.

# Solution

- Combine TypeScript's [Indexed Access Types](https://www.typescriptlang.org/docs/handbook/2/indexed-access-types.html) and [Mapped Types](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html) for better type inference when finding steps.
- Refactor `useSdkOptions()` hook to return a tuple: `[options, enhancedOptions]` which `options` is the original SDK options, while `enhancedOptions` is an object of possible helpers and derived options. In this PR scope: `enhancedOptions.findStep(type: StepTypes)` can be used to find a step config by its type.

## Checklist

_put `n/a` if item is not relevant to PR changes_

- [x] Has the CHANGELOG been updated?
- [n/a] Has the README been updated?
- [n/a] Has the CONTRIBUTING doc been updated?
- [n/a] Has the RELEASE_GUIDELINES been updated?
- [n/a] Has the MIGRATION doc been updated for any MAJOR breaking changes?
- [n/a] Has the MIGRATION doc been updated for any MINOR breaking changes, including any translation strings or keys changes?
- [n/a] Have any new automated tests been implemented or the existing ones changed?
- [n/a] Have any new manual tests been written down or the existing ones changed?
- [n/a] Have any new strings been translated or the existing ones changed?

PR #1416, opened 2021-04-22
Add photoCaptureFallback face step config

natasharsilva:
# Problem
We want integrators to be able to block end-users from uploading selfies if the liveness challenge is absolutely necessary to complete the verification.
We also want end-users on unsupported browser versions to know what they can do next, so that they can try to get verified using Video liveness.

# Solution
Add photoCaptureFallback with default value to true; if the value is set to false, video variant is requested and the browser is not compatible with MediaRecorder, the flow will change as followed:
- if the user is on desktop, it will be redirected to cross-device flow
- if the user is on mobile, it will receive unsupported browser error

If the value is set to true/null, the old flow is used and it should fallback to selfie.

## Checklist


src/components/Capture/withCrossDeviceWhenNoCamera.tsx
```diff
@@ -42,11 +42,12 @@ const withCrossDeviceWhenNoCamera = <P extends CaptureComponentProps>(
         hasCamera,
         requestedVariant,
         step,
-      } = this.props
-
+        photoCaptureFallback,
+      } = { ...this.props, photoCaptureFallback: true }
```
ethan605, line 46:
```suggestion
        photoCaptureFallback = true,
      } = this.props
```

src/components/Router/MainRouter.tsx
```diff
@@ -94,7 +94,24 @@ export default class MainRouter extends Component<InternalRouterProps, State> {
       steps && steps.some(isUploadFallbackOffAndShouldUseCamera)
     const { hasCamera } = this.props
 
-    return !isDesktop && !hasCamera && shouldStrictlyUseCamera === true
+    const faceStep = steps.find(
+      (step) => step.type === 'face'
+    ) as StepConfigFace
```
ethan605, line 99: Since [this PR](https://github.com/onfido/onfido-sdk-ui/pull/1373), we switched to use `buildStepFinder` and `useSdkOptions` hooks for type-guarded step picker from `steps` array.

```suggestion
    const findStep = buildStepFinder(steps)
    const faceStep = findStep('face')
```

Another example for this popular use case: https://github.com/onfido/onfido-sdk-ui/blob/5b8b5ca6533ad58610a946371cb9afac33f3dc1b/src/components/Router/HistoryRouter.tsx#L225-L226

src/components/Router/MainRouter.tsx
```diff
@@ -94,7 +94,24 @@ export default class MainRouter extends Component<InternalRouterProps, State> {
...
+
+    const photoCaptureFallback =
+      faceStep?.options?.photoCaptureFallback == null
+        ? true
+        : faceStep?.options?.photoCaptureFallback
```
ethan605, line 104: You can leverage [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) here

```suggestion
    const photoCaptureFallback = faceStep?.options?.photoCaptureFallback ?? true
```

ethan605 (changes requested):
There're some minor comments.
Also, can you please update the corresponding section in `README.md`, or is this an unofficial change?

src/components/Capture/withCrossDeviceWhenNoCamera.tsx
natasharsilva: Thank you for the review! I'll apply the changes and update the section in the README, should be an official change :)

src/components/Capture/withCrossDeviceWhenNoCamera.tsx
```diff
@@ -42,11 +42,12 @@ const withCrossDeviceWhenNoCamera = <P extends CaptureComponentProps>(
         hasCamera,
         requestedVariant,
         step,
-      } = this.props
-
+        photoCaptureFallback = true,
+      } = { ...this.props }
```
ethan605, line 46: Sorry for being harsh here, but this `{ ...this.props}` will create an unnecessary copy step.

src/components/Router/MainRouter.tsx
```diff
@@ -94,7 +95,20 @@ export default class MainRouter extends Component<InternalRouterProps, State> {
...
+    const photoCaptureFallback = faceStep?.options?.photoCaptureFallback ?? true
+
+    const canVideoFallbackToPhoto =
+      window.MediaRecorder != null || photoCaptureFallback
+
+    return (
+      !isDesktop &&
+      ((!hasCamera && shouldStrictlyUseCamera === true) ||
+        (!canVideoFallbackToPhoto &&
+          faceStep?.options?.requestedVariant === 'video'))
```
ethan605, line 110: I wonder if we can do better here, but not sure how to achieve it. For me, all of these brackets, `&&` and `||` are difficult to understand and maintain in the future. @stefaniacardenas @kopijunkie any suggestions?

src/components/Capture/withCrossDeviceWhenNoCamera.tsx
```diff
@@ -42,11 +42,12 @@ const withCrossDeviceWhenNoCamera = <P extends CaptureComponentProps>(
...
       const currentStep = componentsList[step]
       const cameraRequiredButNoneDetected =
-        !hasCamera &&
+        (!hasCamera ||
+          (window.MediaRecorder == null && !photoCaptureFallback)) &&
```
ethan605, line 50: Ditto here (understandability & maintainability - @stefaniacardenas @kopijunkie)

ethan605 (changes requested):
Overall looks better to me now. Just a couple of concerns about the readability of conditions

src/components/Capture/withCrossDeviceWhenNoCamera.tsx
natasharsilva: no worries! I'll amend the last commit with this update

src/components/Router/MainRouter.tsx
natasharsilva: I agree that this can be hard to read. Not sure how I can make it clearer though. Maybe extracting everything after the `||` to another `const` helps? I'm open to all suggestions 😅

kopijunkie: Yeah, this is quite hard to read. We might be able to make lines 103-104 a util function since it's kind of duplicated in 2 different places/files. 🤔 

But lines 106 onwards is a tough one. Only thing I can think of is break them down into individual constants. 

Not sure if it might be worth creating a util function to check for the required options in the `faceStep` config.

test/specs/scenarios/face.js
```diff
@@ -222,6 +222,24 @@ export const faceScenarios = (lang) => {
...
+        `?language=${lang}&liveness=true&photoCaptureFallback=false`
+      )
+      driver.executeScript('window.MediaRecorder = undefined')
+      documentUpload.clickUploadButton()
+      uploadFileAndClickConfirmButton(
+        passportUploadImageGuide,
+        confirm,
+        'passport.jpg'
+      )
+      await sleep(2500)
```
kopijunkie, line 239: I think in general we want to avoid using `sleep` especially if it's more than 1s or so. We had some issues in the past with flaky tests when using this, if I recall correctly. Have you tried using something like `driver.wait(until.elementIsVisible(crossDeviceIntro.title()), 2500)`?

kopijunkie (changes requested):
Have a suggestion/question for the UI test added & tried to give some suggestion to maybe make the code more readable where the question was raised.

src/components/Router/MainRouter.tsx
natasharsilva: I'll try to make it more readable with those constants and build that util function to see how it looks :)

src/components/Capture/withCrossDeviceWhenNoCamera.tsx
natasharsilva: When removing this copy step I get the following error: `Property 'photoCaptureFallback' does not exist on type 'CaptureComponentProps & Readonly<Attributes & { children?: ComponentChildren; ref?: Ref<any> | undefined; }>'`

I'm very much a beginner in TS, could you help me understand why this might be? I can see that 
```
type CaptureComponentProps =
  | StepComponentDocumentProps
  | (StepComponentFaceProps & { forceCrossDevice: never })
```
So why can't the property that is inside `StepComponentFaceProps` be reached?

test/specs/scenarios/face.js
natasharsilva: this works perfectly and is much better! thanks :)

src/components/Capture/withCrossDeviceWhenNoCamera.tsx
natasharsilva: @ethan605 I updated the type that was throwing the error above and it's now working - but I'm not sure that's the correct way to do this. Do you mind taking a look? 🙏

ethan605: Sorry @natasharsilva, I couldn't take a look at this yesterday. The error was because you were trying to call `photoCaptureFallback` from `this.props`, which could be either `StepComponentDocumentProps` or `StepComponentFaceProps`. But because `photoCaptureFallback` belongs to `StepComponentFaceProps` only, it complained that in some situations when `this.props` has the type of `StepComponentDocumentProps`, then there would be no `photoCaptureFallback` at all.

To fix this, you should change the definition of `CaptureComponentProps` to:

```typescrtipt
type CaptureComponentProps =
  | (StepComponentDocumentProps & { photoCaptureFallback: never })
  | (StepComponentFaceProps & { forceCrossDevice: never })
```

In the latest commit, you put `photoCaptureFallback: true` which is not wrong, but also not reasonable for an absent property at all.

natasharsilva: Makes much more sense now, thanks! I changed it to `never` now along with some changes to try and help with the code readability :)

src/components/Router/StepComponentMap.ts
```diff
@@ -141,12 +141,22 @@ const buildFaceComponents = (
   const shouldSelfieScreenUseCamera =
     !shouldDisplayUploader && deviceHasCameraSupport
 
+  const videoCameraSupport = window.MediaRecorder != null
+
+  const photoCaptureFallback =
+    faceStep?.options?.photoCaptureFallback == null
+      ? true
+      : faceStep?.options?.photoCaptureFallback
```
stefaniacardenas, line 149: This can be simplified to: 

```suggestion
  const photoCaptureFallback = faceStep?.options?.photoCaptureFallback  !== false
```

stefaniacardenas (approved):
LGTM. One minor comment

kopijunkie (approved):

ethan605 (approved):
LGTM!

src/components/Router/StepComponentMap.ts
natasharsilva: updated! :)

MANUAL_REGRESSION.md
```diff
@@ -709,6 +709,17 @@ Given user opened the link with `?uploadFallback=false` flag
    - You should see the document selector screen
 8. You should be able to complete the ID verification flow successfully
 
+### 37. Interrupt flow if selfie fallback is deactivated and MediaRecorder is not supported
+
```
kopijunkie, line 713: Should specify what device OS versions/browsers in general would not have MediaRecorder support & this UX is expected on as Applause testers don't necessarily know this. See earlier Face flow tests in this docs for reference.

natasharsilva: updated! :)

PR #1432, opened 2021-05-11
Task/cx 6461 general refactoring optimisation web sdk pt1

SherBhachuOnFido:
# Problem
- We currently only support Chrome locally
- Inability to select country from dropdown on FF
- Unable to upload documents on Safari locally
- Inaccurate reporting from BrowserStack
- Only able to use a single, hard coded config file

# Solution
- Implement a proper pass/fail solution for BrowserStack executed tests.
- Add Options that allow us to run tests against Safari, Firefox and Edge browsers locally
- Implemented a solution that now allows us to upload files on Safari (both locally and if run on BrowserStack)
- Fixed underlying code that now allows us to select country from dropdown using Firefox
- Added a way to take in custom config file at run time

## Checklist


test/TESTING_GUIDELINES.md
```diff
@@ -10,6 +10,27 @@ Guide to UI tests for Web SDK.
...
+Firefox:
+
+- `npm install` should take care of installing Gekodriver (Webdriver for Firefox)
+- **Note:** if you have OSX Catalina installed, you will need to run `npm install -g geckodriver`, this is due to this
+  issue https://github.com/mozilla/geckodriver/issues/1629
+  <br/>
+
+Safari:
+
+- Simply having Safari installed on your machine is enough<br/>
```
kopijunkie, line 24: Same here. The line break element `<br/>` isn't needed.

test/TESTING_GUIDELINES.md
```diff
@@ -10,6 +10,27 @@ Guide to UI tests for Web SDK.
...
+
+MicroSoft Edge:
+
+1. MicroSoft Edge
```
kopijunkie, line 28: Minor typos here
```suggestion
Microsoft Edge:

1. Microsoft Edge
```

test/TESTING_GUIDELINES.md
```diff
@@ -10,6 +10,27 @@ Guide to UI tests for Web SDK.
...
+2. Microsoft Edge Driver (WedDriver for Edge)
+
+- **Note:** your Microsoft Edge Driver needs to match the Edge browser version you are running and needs to be installed
+  to your path (i.e. on OSX `/usr/local/bin`)
```
kopijunkie, line 32: This can just be on 1 line
```suggestion
- **Note:** your Microsoft Edge Driver needs to match the Edge browser version you are running and needs to be installed to your path (i.e. on OSX `/usr/local/bin`)
```

test/pageobjects/Camera.js
```diff
@@ -1,11 +1,15 @@
...
 import { verifyElementCopy } from '../utils/mochaw'
 import { assert } from 'chai'
+import { browserName } from '../main'
 
 class Camera extends BasePage {
   async enableCameraButton() {
     return this.$('[data-onfido-qa="enable-camera-btn"]')
   }
+  async allowCameraPermissionImage() {
+    return this.$('div.onfido-sdk-ui-CameraPermissions-Primer-graphic')
```
kopijunkie, line 11: Should be able to just specify the class name here.
```suggestion
    return this.$('.onfido-sdk-ui-CameraPermissions-Primer-graphic')
```

test/pageobjects/Camera.js
```diff
@@ -39,6 +43,17 @@ class Camera extends BasePage {
     this.shutterButton().click()
   }
 
+  async enableCameraAccessIfNecessary() {
+    if (browserName === 'safari' || browserName === 'Safari') {
+      if (
+        this.enableCameraButton().isDisplayed() === true &&
+        this.allowCameraPermissionImage().isDisplayed() === true
+      ) {
```
kopijunkie, line 51: Should be enough to just write it like this 
```suggestion
      if (
        this.enableCameraButton().isDisplayed() &&
        this.allowCameraPermissionImage().isDisplayed()
      ) {
```
shorter and also makes it more readable IMO.

test/pageobjects/PassportUploadImageGuide.js
```diff
@@ -50,11 +51,23 @@ export default class PassportUploadImageGuide extends BasePage {
     // Input here cannot use the uploadInput() function above
     const input = this.$('.onfido-sdk-ui-CustomFileInput-input')
     const pathToTestFiles = '../resources/'
-    // This will detect local file, ref: https://www.browserstack.com/automate/node#enhancements-uploads-downloads
-    this.driver.setFileDetector(new remote.FileDetector())
-    const sendKeysToElement = input.sendKeys(
-      path.join(__dirname, pathToTestFiles + filename)
-    )
+    //the below line...if safari AND Local, ignore it...otherwise file uploads will fail.
+    if (browserName === 'safari' && isRemoteBrowser === false) {
```
kopijunkie, line 55: I noticed in the [`enableCameraAccessIfNecessary`](https://github.com/onfido/onfido-sdk-ui/blob/ccd5551b54274e8df910a9d31ba4ee7dd774ceb9/test/pageobjects/Camera.js#L47) function you are checking for both `browserName === 'safari' || browserName === 'Safari'` but the 2nd `browserName` case isn't required here also?

test/pageobjects/Welcome.js
```diff
@@ -51,6 +51,7 @@ class Welcome extends BasePage {
   }
 
   async continueToNextStep() {
+    this.driver.sleep(100)
```
kopijunkie, line 54: I think there should be a note here about why this explicit `sleep` is needed here.

test/specs/scenarios/crossDevice.js
```diff
@@ -215,6 +211,7 @@ export const crossDeviceScenarios = async (lang) => {
 
       it('should send sms and navigate to "Check your mobile" screen ', async () => {
         driver.get(baseUrl)
+        driver.navigate().refresh()
```
kopijunkie, line 214: Why is this browser refresh needed here in this test?

kopijunkie (changes requested):
Nice job! Mostly questions & some minor nitpicks. Keen to see how these tests run once our CI/CD builds are back to normal working order. 😄

test/pageobjects/PassportUploadImageGuide.js
SherBhachuOnFido: Yes thats correct. The way Selenium reads the name is like this... 'safari' is local and 'Safari' is browserstack.

test/specs/scenarios/crossDevice.js
SherBhachuOnFido: For some reason, on safari, I had noticed if this was not there, it would never pass.

test/config.json
```diff
@@ -16,19 +16,21 @@
...
           "os_version": "10",
           "resolution": "1920x1080",
-          "remote": true
+          "remote": true,
+          "name": "multipleBrowsersDocumentUpload"
         },
         {
           "browserName": "Edge",
-          "browser_version": "87",
+          "browser_version": "89",
```
kopijunkie, line 28: Maybe I remembered it wrongly, but we can't set `"browser_version"` to `"latest"` for Edge also like for Firefox?

kopijunkie (approved):
1 minor typo. & a question. Otherwise looks good.

test/config.json
SherBhachuOnFido: We can but I noticed that setting it to latest it was throwing some exceptions, basically it looks like there is an issue on BrowserStack side where they may be using an incorrect version of the MSEdgeDriver on their side, so I didnt want us to unnecessarily see exceptions on our side.

kopijunkie: Ah, ok. Makes sense then to not use that setting.

kopijunkie (approved):

test/TESTING_GUIDELINES.md
```diff
@@ -10,6 +10,25 @@ Guide to UI tests for Web SDK.
...
+- **Note:** if you have macOS Catalina installed, you will need to run `npm install -g geckodriver`, this is due to this
+  issue https://github.com/mozilla/geckodriver/issues/1629
+
+Safari:
+
+- Simply having Safari installed on your machine is enough
+
+Microsoft Edge:
+
+1. Microsoft Edge
```
stefaniacardenas, line 27: Any specific version? Intel vs Apple chip?

test/TESTING_GUIDELINES.md
```diff
@@ -10,6 +10,25 @@ Guide to UI tests for Web SDK.
...
+2. Microsoft Edge Driver (WebDriver for Edge)
```
stefaniacardenas, line 28: Might be good to include a link or command to make developers' lives easier :)

test/TESTING_GUIDELINES.md
```diff
@@ -25,6 +44,17 @@ Our UI tests also run on BrowserStack. If you would like to run them on your Bro
...
+
+- Within `config.json` you simply need to modify the `browserName` value to any of the ones illustrated
+
+`"browsers": [ { "browserName": "chrome | firefox | safari | MicrosoftEdge", "remote": false }`
+
+### BrowserStack notes
+
+- Specific to Safari on BrowserStack, you need to explicitly set the `localhostUrl` value in `config.json` to
+  be read as `"localhostUrl": "https://bs-local.com:8080/",` this is due to the way Safari handles `https://localhost`.
+
```
stefaniacardenas, line 57: I think it would be worth including these configuration in `config.json` but only execute them if we pass flag ie. `npm run test:ui --browser safari`

test/main.js
```diff
@@ -133,6 +170,37 @@ const createMocha = (driver, testCase) => {
   // By default `require` caches files, making it impossible to require the same file multiple times.
   // Since we want to execute the same tests against many browsers we need to prevent this behaviour by
   // clearing the require cache.
+  mocha.suite.beforeAll('Do something before everything', function () {
```
stefaniacardenas, line 173: Might be worth adding a more meaningful message here
```suggestion
  mocha.suite.beforeAll('Setting up browser before tests execution, function () {
```

test/main.js
```diff
@@ -133,6 +170,37 @@ const createMocha = (driver, testCase) => {
...
+    global.isRemoteBrowser = isRemoteBrowser
+    global.browserName = browserName
+  })
+  mocha.suite.beforeEach('Set retry', function () {
+    this.currentTest.retries(2)
+  })
+  mocha.suite.afterEach('Capture total number of test failures', function () {
+    const currentTestState = this.currentTest.state
+    //console.log(`The current test state is: ${currentTestState}`)
```
stefaniacardenas, line 182: You might have forgot this log here

test/pageobjects/DocumentUpload.js
```diff
@@ -32,11 +33,23 @@ class DocumentUpload extends BasePage {
     // Input here cannot use the uploadInput() function above
     const input = this.$('.onfido-sdk-ui-CustomFileInput-input')
     const pathToTestFiles = '../resources/'
-    // This will detect local file, ref: https://www.browserstack.com/automate/node#enhancements-uploads-downloads
-    this.driver.setFileDetector(new remote.FileDetector())
-    const sendKeysToElement = input.sendKeys(
-      path.join(__dirname, pathToTestFiles + filename)
-    )
+    //the below line...if safari AND Local, ignore it...otherwise file uploads will fail.
+    if (browserName === 'safari' && isRemoteBrowser === false) {
```
stefaniacardenas, line 37: Can `isRemoteBrowser` be undefined or null?

test/pageobjects/PassportUploadImageGuide.js
```diff
@@ -50,11 +51,23 @@ export default class PassportUploadImageGuide extends BasePage {
...
+      this.driver.setFileDetector(new remote.FileDetector())
+    }
+    const filePath = path.join(__dirname, pathToTestFiles + filename)
+    let sendKeysToElement
+    try {
+      console.log(`File being uploaded is ${filePath}`)
+      sendKeysToElement = input.sendKeys(path.join(filePath))
+    } catch (err) {
+      console.log('Just logging that I have caught an exception on upload')
+    }
```
stefaniacardenas, line 70: This could be a utils function and be used here and in `DocumentUpload.js`

stefaniacardenas (changes requested):
It's looking good! I left some comments

test/TESTING_GUIDELINES.md
SherBhachuOnFido: Nothing specific, this isn't like x86/M1, this is the same as Chrome/Chromedriver, i.e. if you have v90 of Chrome, you need the matching v90 of Chromedriver, if that makes sense.

test/TESTING_GUIDELINES.md
SherBhachuOnFido: Ah, will do.

SherBhachuOnFido: Will do.

test/main.js
SherBhachuOnFido: That was actually just for my debugging, its not needed.

test/pageobjects/PassportUploadImageGuide.js
SherBhachuOnFido: Cool, I'll take a look at refactoring that.

test/pageobjects/DocumentUpload.js
SherBhachuOnFido: Nope, it cant be, it has to be explicitly set in the config...(take a look at config.json)

"remote": false or "remote": true

test/TESTING_GUIDELINES.md
SherBhachuOnFido: Well the thing is, the idea here is not to execute based on Browser, but config file. This is the next step because I didn't want to create all the config files with different settings without us having an agreed upon approach.

I can certainly implement the use of an environment variable, which would be my preference because if i pass in `--browser safari` ... are we implying local safari or BS safari? Also, would we choose a default browser if one is not specified? I'm just a little hesitant to add something like this now. My preference would be a specific config file then have the browser you want to run within that as a parameter.

stefaniacardenas: I see. I agree, this can be handled as part of another story

stefaniacardenas (approved):

test/TESTING_GUIDELINES.md
SherBhachuOnFido: Great. I think we already have a ticket for this in the backlog.

PR #1460, opened 2021-05-26
Task/Update welcome screen UI

kopijunkie:
# Task
The end user should see the new layout for the Welcome screen (see design screenshot below) 

Screenshot 2021-05-26 at 19 32 26

- By default, user should see the default instructions as per Figma design

- Depending on the SDK configuration user should see either document, face or document + face in the instruction text 

- If custom `descriptions` copy for the Welcome screen is supplied by the integrator, the user should see the custom descriptions instead of the default instructions

## Checklist

- [x] Has the CONTRIBUTING doc been updated?


stefaniacardenas (approved):
Looks good. Do we need to add the strings that changed to the migration file?

src/components/Welcome/index.tsx
```diff
@@ -68,24 +70,41 @@ const WelcomeActions: FunctionComponent<WelcomeActionsProps> = ({
...
+    : ''
 
-  const actions = <WelcomeActions {...{ nextButton, nextStep }} />
-  const welcomeTitle = title ? title : translate('welcome.title')
+  const configuredIDVSteps = steps
+    .filter((step) => step.type === 'document' || step.type === 'face')
+    .map((stepConfig) => stepConfig.type)
+  const welcomeDescriptions = customDescriptions
+    ? customDescriptions
+    : getLocalisedDescriptions(configuredIDVSteps, translate)
```
stefaniacardenas, line 102: What will this display for `poa` or `auth`? Shouldn't we have a default for those cases?

stefaniacardenas (changes requested):

src/components/Welcome/index.tsx
kopijunkie: Good point. I'm not sure about messaging for `auth` or how to preview it so will need to catch up with @carlagn but for `poa` we do need some kind of default messaging in English.

kopijunkie:
> Looks good. Do we need to add the strings that changed to the migration file?

I'm conflicted about whether to add this now or later when all the copy update happens. But could also be worth getting started now.

src/components/Welcome/index.tsx
kopijunkie: I've mapped the `auth` Welcome screen copy to `face` for now. Auth is currently just using the existing SDK's welcome screen but will bring it up on Slack for discussion with designers.

src/locales/de_DE/de_DE.json
```diff
@@ -476,9 +476,11 @@
...
     "welcome": {
-        "description_p_1": "Um ein Bankkonto zu eröffnen, müssen wir Ihre Identität überprüfen.",
-        "description_p_2": "Es wird nur ein paar Minuten dauern.",
         "next_button": "Identität überprüfen",
-        "title": "Eröffnen Sie Ihr neues Bankkonto"
+        "title": "Identität überprüfen",
+        "subtitle": "It should take a few minutes",
+        "description_p_1": "Use your webcam or phone to photograph:",
+        "description_p_2": "1. your identity document",
+        "description_p_3": "2. your face"
```
stefaniacardenas, line 484: Why is this partially translated? Same for the other languages

kopijunkie: I can update them with the latest from Lokalise now, but the description keys all have the localised translations for "Verify Identity". Have raised this question in Slack about the translations.

kopijunkie: Updated to use the latest key-string for default descriptions after discussion with Rae.

src/demo/SdkMount.tsx
```diff
@@ -17,6 +17,10 @@ import * as Onfido from '../index'
 */
 
 const Onfido = window.Onfido
```
stefaniacardenas, line 19: For Auth you should be able to just do `const Onfido = window.OnfidoAuth` which works fine in the sample app but for some reason it doesn't in this demo app

stefaniacardenas (approved):

src/demo/SdkMount.tsx
kopijunkie: Interesting. I tried `import * as Onfido from '../index'` after talking to Carla about how to get it running locally & she suggested this, which worked. So just went with that.

ethan605 (changes requested):
Just some minor changes needed

src/demo/demoUtils.ts
```diff
@@ -147,7 +148,22 @@ export const getInitSdkOptions = (): SdkOptions => {
...
+              title: 'Open your new bank account',
+              descriptions: [
+                'To open a bank account, we will need to verify your identity.',
+                'It will only take a couple of minutes.',
+              ],
+              nextButton: 'Verify Identity',
+            },
+          },
+        ]
+      : [{ type: 'welcome' }]
```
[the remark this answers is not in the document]
kopijunkie: Agreed. Looking at the comparison here, the IF-ELSE is easier to read/follow.

ethan605 (approved):

PR #1470, opened 2021-06-08
Bugfix/country selection skipped unexpectedly

### LINKED ISSUE #1444: Country selection is skipped unexpectedly
### What was the expected behaviour?

Given the options `documentTypes: { driving_licence: true, national_identity_card: true}` I expected to see the country-selection screen.

### What happened instead?

The country-selection screen was skipped (as if I had set the country property to `{country:null}`).
However setting this property to any other value makes the selection screen appear again.

### Version info:

- 6.7.0/6.8.0
- Linux
- Firefox
- Node 12

### Steps to reproduce:

Start the Onfido document check and click on "driver license".

kopijunkie:
# Problem
From Web SDK v6.6.0 integrations that have enabled document type options using the boolean configuration option skips the Country Selection screen instead of displaying like it was in previous versions as of v6.3.1. Fixes #1444 

# Solution
- Add a conditional check for when a document type's configuration is just a boolean value
- Add UI tests to ensure this configuration's UX works as intended

## Checklist


src/components/Router/StepComponentMap.ts
```diff
@@ -267,7 +267,9 @@ const buildDocumentComponents = (
   )
   const showCountrySelection =
     showCountrySelectionForSinglePreselectedDocument ||
-    (!hasOnePreselectedDocument && !supportedCountry && countryCode !== null)
+    (!hasOnePreselectedDocument &&
+      !supportedCountry &&
+      (countryCode !== null || configForDocumentType === true))
```
ethan605, line 272: I believe this could be broken into smaller & more verbose checks :thinking:

src/demo/demoUtils.ts
```diff
@@ -25,6 +25,7 @@ export type QueryParams = {
   liveness?: StringifiedBoolean
   multiDocWithInvalidPresetCountry?: StringifiedBoolean
   multiDocWithPresetCountry?: StringifiedBoolean
+  multiDocBooleanWithCountrySelection?: StringifiedBoolean
```
ethan605, line 28: IMO this could be
```suggestion
  multiDocWithBooleanValues?: StringifiedBoolean
```

src/demo/demoUtils.ts
```diff
@@ -130,6 +131,13 @@ const getPreselectedDocumentTypes = (): Partial<
     }
   }
 
+  if (queryParamToValueString.multiDocBooleanWithCountrySelection === 'true') {
```
ethan605, line 134:
```suggestion
  if (queryParamToValueString.multiDocWithBooleanValues === 'true') {
```

ethan605 (approved):
LGTM!

src/components/Router/StepComponentMap.ts
kopijunkie: Agree. I've taken another stab at it, but if you have any ideas for further improvements, feel free to suggest.

src/components/Router/StepComponentMap.ts
```diff
@@ -265,9 +265,14 @@ const buildDocumentComponents = (
     countryCode,
     documentType
   )
+  const hasMultipleDocumentsWithUnsupportedCountry =
+    !hasOnePreselectedDocument && !supportedCountry
+  const hasCountryCodeNullOrBooleanDocumentTypeFlag =
```
stefaniacardenas, line 270: `hasCountryCodeNullOrBooleanDocumentTypeFlag` should probably be renamed to `hasCountryCodeOrDocumentTypeFlag` because in the code you are checking that `countryCode` is not `null`

src/components/Router/StepComponentMap.ts
```diff
@@ -265,9 +265,14 @@ const buildDocumentComponents = (
...
+    countryCode !== null || configForDocumentType === true
```
stefaniacardenas, line 271: Why not 
```suggestion
    countryCode !== null || configForDocumentType
```

src/components/Router/StepComponentMap.ts
kopijunkie: Good point. Will rename it.

src/components/Router/StepComponentMap.ts
kopijunkie: Actually, realised this would need to be a strict check for a boolean `true` value since `configForDocumentType` could also be an object which is truthy also.

stefaniacardenas (approved):

PR #1477, opened 2021-06-15
Bugfix/live doc capture camera display

kopijunkie:
# Problem
When Document Live Capture feature is enabled, on some devices and with device emulators on Chrome the camera view is out of position relative to the document frame overlay. This affects the document capture preview in the Confirm Upload screen.

# Solution
Make camera view line up with the overlay based on device viewport/pixel density.

## Checklist


src/components/Camera/style.scss
```diff
@@ -112,8 +112,33 @@
...
+
+    @media (orientation: portrait) {
+      height: 100%;
+      width: 100%;
+    }
+
+    @media (orientation: landscape) {
+      height: auto;
+      width: 100%;
+      top: 21.5%;
```
stefaniacardenas, line 125: How did you get to this value?

src/components/Camera/style.scss
```diff
@@ -112,8 +112,33 @@
...
+    }
+
+    @media (--small-viewport) and (--below-high-pixel-density) {
+      height: auto;
+      width: 100%;
+      top: -34%;
```
stefaniacardenas, line 131: How did you get to this value?

src/components/Theme/custom-media.css
```diff
@@ -7,3 +7,7 @@
 @custom-media --short-viewport (max-height: 567px);
 /* For viewport larger than iPhone 5 in terms of height */
 @custom-media --long-viewport (min-height: 568px);
+
+@custom-media --high-pixel-density (-webkit-min-device-pixel-ratio: 3);
+
+@custom-media --below-high-pixel-density (-webkit-max-device-pixel-ratio: 2);
```
ethan605, line 13: Can we have more explanation here? E.g. which devices come with `--high-pixel-density` and which come with `--below-high-pixel-density`?

ethan605 (approved):
Generally looks good to me

src/components/Camera/style.scss
```diff
@@ -112,8 +112,33 @@
...
+      margin: 50% auto;
+    }
+
+    @media (--high-pixel-density) and (orientation: portrait) {
+      top: -20%;
+    }
+
+    @media (--high-pixel-density) and (orientation: landscape) {
+      top: 0;
+    }
```
stefaniacardenas, line 141: Why do we need to do all this when for selfie we can just use `object-fit: cover`?

src/components/Camera/style.scss
kopijunkie: For both instances, through trial & error with different devices. In this case, starting with the default for when the modal is not taking the full width, height of a device (tablets).

src/components/Camera/style.scss
kopijunkie: Good question. The implementation used to be just using `object-fit: cover` similar to selfie but was changed earlier this year to fix an issue with zoomed in live document capture view on the higher spec Huawei phones.

stefaniacardenas: I wonder why those Huawei phones don't have the zooming issue on selfie. Maybe because of multiple environmental cameras?
Anyways, I can see that the live document capture is still broken on desktop and although the feature is not meant to be used on desktop, we can still fix it by using `object-fit: cover` for desktop only?

src/components/Camera/style.scss
kopijunkie: Similar answer as above.

src/components/Camera/style.scss
kopijunkie: Yeah, I think so too (re: Huawei phones). Since only thing I can think of is the selfie isn't using the same back camera. 

Will add the adjustment to use `object-fit: cover` for desktop only. 👍

stefaniacardenas (approved):

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ This project adheres to the Node [default version scheme](https://docs.npmjs.com
 - Internal: Upgrade Preact from version `10.5.4` to `10.5.13` in order to resolve an unhandled exception on reinitialising the SDK after closing the SDK modal for some integrations when using Modal mode.
 - Public: Fix Country Selection screen not displaying when SDK is initialised with boolean `documentTypes` configuration.
 - Public: Fixed issue where `onComplete` callback was fired without the necessary data for the `face` step. Added exception to `onError` callback to inform that `onComplete` could not be fired due to missing data.
+- UI: Fix camera view not lining up with Document Live Capture overlay.
```
stefaniacardenas, line 43: This should go under `next-version`

src/components/Camera/index.tsx
```diff
@@ -16,6 +16,7 @@ import type {
   WithTrackingProps,
   WithPermissionsFlowProps,
 } from '~types/hocs'
+import { isDesktop } from '~utils'
```
stefaniacardenas, line 19: Looks like this is not used

stefaniacardenas (changes requested):

stefaniacardenas (approved):
LGTM!

ethan605 (approved):
LGTM!

PR #1489, opened 2021-06-29
Set of E2E Tests against Production API

SherBhachuOnFido:
# Problem
As part of a post mortem, it was identified that we could potentially have caught a live issue if we have e2e tests running after we'd released.

# Solution
Implement a set of e2e tests to run against production. This initial set will run on Chrome, Safari and MS Edge. FF Has proved to be problematic and IE will require some specific handling, so I shall work on those in a follow up.

I have implemented a new GitHub workflow that explicitly uses a ```macos-latest``` host so that all the above browsers can work, rather than rely on BS.

## Checklist

- [x] Has the TESTING_GUIDELINES been updated?


test/main.js
```diff
@@ -17,12 +17,15 @@ import edge from 'selenium-webdriver/edge'
 let config
 const browsersFailures = {}
 let totalFailures = 0
+export let localhostUrl
 
 if (!process.env.CONFIG_FILE) {
   console.error('INFO: CONFIG_FILE not set, so using the default config.json')
   config = require('./config')
+  localhostUrl = config.localhostUrl
```
kopijunkie, line 25: IMO it's less confusing to assign specific variable names for each of these config instances rather than use a generic `config` variable name. So in this case, I would do this:
```suggestion
  const defaultConfig = require('./config')
  localhostUrl = defaultConfig.localhostUrl
```

test/main.js
```diff
@@ -17,12 +17,15 @@ import edge from 'selenium-webdriver/edge'
...
 } else {
   config = require(`./${process.env.CONFIG_FILE}`)
+  localhostUrl = config.localhostUrl
```
kopijunkie, line 28: And here I would do this: 
```suggestion
  const envConfig = require(`./${process.env.CONFIG_FILE}`)
  localhostUrl = envConfig.localhostUrl
```

test/main.js
```diff
@@ -395,7 +408,11 @@ const runMockServerAndTests = () =>
     findMockServerId(runTests)
   })
 
-runMockServerAndTests()
-
+if (process.env.MOCK_SERVER === 'false') {
+  console.log('Tests will run without the use of the mock server')
```
kopijunkie, line 412: Better to explicitly state that the tests will be running against production to be clear

test/specs/scenarios/face.js
```diff
@@ -340,7 +340,7 @@ export const faceScenarios = (lang) => {
       faceVideoIntro.verifyUIElementsOnTheFaceVideoIntroScreen(copy)
       faceVideoIntro.clickOnContinueButton()
       camera.enableCameraButton().click()
-      camera.verifyVideoTitle(copy)
+      //camera.verifyVideoTitle(copy) - copy difference
```
kopijunkie, line 343: What was the copy difference here?

kopijunkie (changes requested):

.github/workflows/e2e_tests.yml
```diff
@@ -0,0 +1,28 @@
...
+  pull_request:
+
+jobs:
+  build:
+    runs-on: macos-latest
+
+    steps:
+      - uses: actions/checkout@v2.3.4
+      # Needed to be able to run `npm ci`. It fails if we try to install libraries from a github repo
+      # e.g. "wpt": "git+https://github.com/Woopra/browser-tracker.git#v5.1.20",
```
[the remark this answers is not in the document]
SherBhachuOnFido: I can remove this, as I copied/pated the content originally from an existing GH .yml file

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ This project adheres to the Node [default version scheme](https://docs.npmjs.com
 - UI: Accessibility - Add ARIA role `button` to "Resend link" and "Cancel" links on Cross Device flow SMS Sent, Mobile Connected screens.
 - Internal: Remove `image_quality` breakdowns from `sdk_warnings` response because the field will be soon deprecated.
 - Internal: Moved `geckodriver` from `dependencies` to `devDependencies`.
+- Internal: Added ability to run tests without the use of the mock server.
```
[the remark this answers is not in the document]
SherBhachuOnFido: Good point! will do.

test/main.js
SherBhachuOnFido: Noted, I'll make that change.

test/main.js
SherBhachuOnFido: Noted, I'll make that change.

test/main.js
SherBhachuOnFido: Noted, I'll make that change.

test/main.js
SherBhachuOnFido: Ok, there is a problem here, it comes down to the way the test runner is setup, it uses the variable ```config```, so using a ```defaultConfig``` and ```envConfig``` will require this to be re-written and I really dont want to touch that thing at all.

test/specs/scenarios/face.js
SherBhachuOnFido: Oh, I just ran it uncommented, it looks ok now, there might have been a copy issue when I first created the branch.

kopijunkie (approved):
LGTM

PR #1491, opened 2021-06-30
Some users blocked from completing IDV flow at Face liveness video upload step

kopijunkie:
# Problem
Users on some mobile devices, desktops are getting blocked from completing IDV flow at the Face liveness video upload step due to the camera falling back to the browser's default video recording format when WebM format is not supported. This could result in a larger than expected video file, e.g. a 12s 1080p MP4 video file is over 10MB in size. 

API does not accept files over 10MB in size and results in users seeing the generic "Connection lost" error. But a 10MB file also takes over 1-2mins to upload on a slow/bad connection so users either think they are stuck on the SDK's loading spinner screen, give up on the upload or the upload request times out before the error can be returned.

# Solution
Reduce resolution in video settings to VGA if browser does not support recording a video in WebM format to avoid generating a large video file.

## Checklist


src/components/Camera/index.tsx
```diff
@@ -53,16 +58,29 @@ const Camera: FunctionComponent<Props> = ({
...
+  // Specify just a camera width (no height) because on safari if you specify both
+  // height and width you will hit an OverconstrainedError if the camera does not
+  // support the precise resolution.
+  // Using width here because on some special devices (e.g. Samsung Galaxy),
+  // setting 720px in height results to 720x960 resolution instead of the desired 720x1280.
+  //
+  // Resolution needs to be set to a lower value if WebM format is not supported
+  // as video formats like MP4 (Safari) result in larger video file sizes
+  // * Resolutions: 1280px = 720p, 480px = VGA (VGA is minimum we can go for automation + iOS is using VGA resolution)
+  const defaultCameraWidthInPx = isWebmFormatSupported() ? 1280 : 480
```
ethan605, line 70: Is `480` too low? Can we use something higher (like `560` or `720`)?

src/types/commons.ts
```diff
@@ -57,6 +57,12 @@ export type SdkMetadata = {
     browser: string
     browser_version: string
   }
+  camera_settings?: {
+    aspect_ratio: number | undefined
+    frame_rate: number | undefined
+    height: number | undefined
+    width: number | undefined
```
ethan605, line 64: I believe these could be shortened:

```suggestion
    aspect_ratio?: number
    frame_rate?: number
    height?: number
    width?: number
```

ethan605 (changes requested):
LGTM, except the fallback resolution :pray:

src/components/Camera/index.tsx
kopijunkie: Discussed with Bio team & it seems that iOS SDK was using way lower resolution (VGA) setting than Web. But I'll do some tests to see how big the file sizes are at `560` in case that's viable. I tried `720` and that did still result in some of the over 10s videos being over 10MB in size.

ethan605 (approved):
LGTM!

carlagn (approved):
LGTM:)

PR #1521, opened 2021-07-28
Add onUpload to capture document uploads

lf94:
After a document upload it's necessary for some companies to keep a copy of the
uploaded document. The Onfido component already returns this information from
a successful POST /documents, but does not expose it in any way. Developers can
now optionally pass an `onUpload` property to access this data.

I did this PR as quickly as possible, and at the very least, made sure everything type-checked. I spent a total of maybe 30
minutes to 1. learn the codebase as I navigated it 2. implement the solution.

If this PR is not merged, BitySA will host these changes as they are vital to our business operations, and may be located at @bity/onfido-sdk-ui.

Thanks!


lf94:
@ethan605 I'm under the impression that this PR is not being noticed (probably because another system is being used to interact with GitHub). Ping :slightly_smiling_face:

ethan605:
Hi @lf94, thank you for submitting this PR. Recently we're processing multiple internal features so please bear with us for the late response. I'll take a look at this tomorrow and get back to you with our final answer for this feature.

src/components/Confirm/Confirm.js
```diff
@@ -112,9 +112,11 @@ class Confirm extends Component {
   }
 
   onApiSuccess = (apiResponse) => {
-    const { method, nextStep, actions } = this.props
+    const { method, nextStep, actions, triggerOnUpload = () => {} } = this.props
     const { capture } = this.state
 
+    triggerOnUpload(apiResponse);
```
ethan605, line 118: Hi @lf94, as what I'm understanding at this point, you want to save the `apiResponse`, which is a JSON of several fields with `applicant_id`, document `id` and the 2 hrefs with document `id`. I believe that:
- You have the full access to `applicant_id`
- You have the necessary document `id` emitted via [`onComplete` callback](https://github.com/onfido/onfido-sdk-ui/#oncomplete-function-optional)

Can you please explain more about your need for such a dedicated `onUpload` callback?

lf94: Yes, it's possible to get all the information at some point in time using multiple requests. The `onUpload` resolves the following:

* No extra requests to get the information needed to download the documents uploaded to Onfido
* Specifically fires after upload of a single document

The `onComplete` callback "fires when both the document and face have been successfully captured and uploaded". 
Waiting for a face upload makes sense when using `onComplete`, but it doesn't make sense when a developer wants to capture documents as they are uploaded.

lf94: As it's been some time, we've decided to use the webhook instead, so this callback is no longer required for us, but it may be for someone else.

ethan605: > we've decided to use the webhook instead, so this callback is no longer required for us

I'm glad that you've figured out an alternative way to make this work instead of self-hosting a patched version of Onfido Web SDK.

> but it may be for someone else. I can image such a feature can be used for UX.

Yes, we always appreciate the contributions towards improving UX for the Web SDK. But IMHO this feature will involve more than just UX because it relates to the Web SDK design principles, which focus on providing a consistent & robust experience in media capturing for the end-users, not media retrieving for the integrators. I believe that the Web SDK should be a black box, receiving minimum inputs (the SDK token) and returning minimum outputs (the document & biometric IDs).

But I'll pass your idea to our team's discussions to see if we can fit it into our product roadmap.

Thank you for your contribution!

lf94: >  I believe that the Web SDK should be a black box

If that's one of the goals then I totally agree with your assessment and also believe the feature is "too low level". :slightly_smiling_face: 

Thanks again Ethan.

PR #1522, opened 2021-07-30
Task/cx 6763 add additional languages to percy

SherBhachuOnFido:
# Problem
We currently only have tests on Percy running against English, which may not capture potential issues.

# Solution
Add additional languages that may capture issues going forward...
- Added es_ES.

## Checklist


test/specs/scenarios/proofOfAddress.js
```diff
@@ -251,15 +251,18 @@ export const proofOfAddressScenarios = async (lang = 'en_US') => {
         poaDocumentSelection.clickOnBankIcon()
         poaGuidance.clickOnContinueButton()
         documentUpload.switchToCrossDevice()
-        await takePercySnapshot(driver, 'Continue on your phone screen for PoA')
+        await takePercySnapshot(
+          driver,
+          `Continue on your phone screen for PoA in ${lang}`
+        )
```
kopijunkie, line 257: PoA is only supported for English so it's not necessary to take a screenshot for it in other languages. 

I'm also reasonably sure there is no difference between Cross Device Intro screens, i.e. the "Continue on your phone screen", for PoA vs the regular IDV flows. 🤔

kopijunkie (changes requested):

test/specs/scenarios/proofOfAddress.js
SherBhachuOnFido: Do you think we should consider removing this specific capture totally then?

kopijunkie: Yeah, I think we should remove it here. I believe this screen's snapshot is already covered in another UI test scenario anyway.

test/specs/scenarios/proofOfAddress.js
```diff
@@ -251,15 +251,18 @@ export const proofOfAddressScenarios = async (lang = 'en_US') => {
...
         crossDeviceIntro.continueToNextStep()
         assert.isTrue(
           crossDeviceLink.qrCode().isDisplayed(),
           'Test Failed: QR Code should be visible'
         )
         await takePercySnapshot(
           driver,
-          'Get your secure link screen - QR Code',
+          `Get your secure link screen - QR Code in ${lang}`,
```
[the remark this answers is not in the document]
SherBhachuOnFido: Do you mean that we only need to test this in English?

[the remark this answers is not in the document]
kopijunkie: Ah, no. I mean in that it's no different on PoA flow as on regular flow. I happened to have added screenshots for the Cross Device Link screen(s) under the [`crossDevice`](https://github.com/onfido/onfido-sdk-ui/blob/07458e8cb445f32d3b35b7ed62bac7bacbc66653/test/specs/scenarios/crossDevice.js) scenario when I worked on PR #1528 so there's probably a slight repetition of work here. But IMO that was a better place for it.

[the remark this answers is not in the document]
SherBhachuOnFido: Ah ok, so remove this one too you mean.

test/percy.json
```diff
@@ -13,6 +13,13 @@
...
   "localhostUrl": "https://localhost:8080/",
   "supportedLanguages": [],
-  "fullTestCoverageLanguages": ["en_US"],
+  "fullTestCoverageLanguages": [
+    "en_US",
+    "es_ES",
+    "de_DE",
+    "fr_FR",
+    "it_IT",
+    "pt_PT"
```
stefaniacardenas, line 22: Just want to confirm that this is not slowing down the build. Is it running in parallel?

SherBhachuOnFido: They won't run in parallel, since the way the tests are expected to run are they iterate through each language. So, it was taking apprx 5mins originally and now takes approx 25mins. 

There is some documentation around parallel execution for Percy here : https://docs.percy.io/docs/parallel-test-suites.

I am happy to create a separate percy profile to run these scenarios on an on-demand or overnight/weekly basis. We can also leave this in draft until we have a working parallel solution working.

stefaniacardenas: I think it would be good to run this overnight against development branch

kopijunkie: Agree about this being run overnight instead if it's going to have that big of an impact on the build time.

SherBhachuOnFido: Sure, makes sense. Let me figure out how best to do this, then I can at least update the PR. I think the best solution might be to leave what we currently have in the 'as-is' Percy setup, which seems to be working quite well and create a new config.

kopijunkie (approved):
LGTM. Not sure why the action doesn't seem to run still. Maybe because the build failed? 🤔

SherBhachuOnFido:
@kopijunkie I don't really think we'd need to add this to the Changelog, @stefaniacardenas what do you think?

stefaniacardenas:
You could add it as an internal entry as a reference for contributors, but I think it's fine to merge this without a Changelog entry.

SherBhachuOnFido:
Cool, I've still not managed to figure out why the CRON is not running, thats the last remaining bit, but seeing as its low priority, we can perhaps take a look later next week once Im back.

stefaniacardenas (approved):
Given we are merging this into master, I think it now makes sense to add a Changelog entry.

stefaniacardenas (approved):

PR #1528, opened 2021-08-09
Improvement/specify available cross device link options

kopijunkie:
# Problem
Some integrators would prefer to have only the SMS option or limit options to just 2 options, e.g. SMS and copy link only, for their end users rather than place focus on the QR code option for various reasons. 

# Solution

- allow integrators to pass an array of methods in the order of preference i.e. `_crossDeviceLinkMethods: ['qr_code', 'copy_link']`
    - note underscore prefix to indicate it is an undocumented feature
    - 1st option is always the default initial view
    - if array only contains 1 option, e.g. `_crossDeviceLinkMethods: ['sms']`, UI will adjust & not show the "or choose one of these alternatives" section
    - if array is empty or has invalid values, SDK should fall back to how it is by default now and show a console warning that there must be at least 1 option provided
- Also fixed an issue with the QR code not rendering fully on initial screen load on some Safari desktop browsers 

## Checklist


src/components/crossDevice/CrossDeviceLink/index.js
```diff
@@ -38,6 +38,24 @@ const SECURE_LINK_VIEWS = [
   },
 ]
 
+const configHasInvalidViewIds = (viewIdsInConfig) => {
+  const validViewIds = SECURE_LINK_VIEWS.map(({ id }) => id)
+  const invalidViewIds = viewIdsInConfig.filter(
+    (viewId) => !validViewIds.includes(viewId)
+  )
```
rabelloo, line 45: 💅🏻  [nitpick]

Probably does not have a huge runtime impact, but the time complexity could be reduced by using a Set rather than an array.

```suggestion
  const validViewIds = new Set(SECURE_LINK_VIEWS.map(({ id }) => id))
  const invalidViewIds = viewIdsInConfig.filter(
    (viewId) => !validViewIds.has(viewId)
  )
```

src/components/crossDevice/CrossDeviceLink/index.js
```diff
@@ -38,6 +38,24 @@ const SECURE_LINK_VIEWS = [
...
+  if (invalidViewIds.length > 0) {
+    console.log(
+      'WARNING - Invalid properties in _crossDeviceLinkMethods option:',
+      invalidViewIds.join(', ')
+    )
+    console.log(
+      '_crossDeviceLinkMethods must be an array with at least 1 of the following option: "qr_code", "copy_link", "sms"'
+    )
```
rabelloo, line 53: If these are warnings, should they be `console.warn` instead?
Or perhaps should they simply `throw`, since it'd be an error they can notice and recover from during development?
I can't imagine a scenario where it'd work in development but throw in production 🤔

src/components/crossDevice/CrossDeviceLink/index.js
```diff
@@ -307,18 +327,48 @@ class CrossDeviceLinkUI extends Component {
...
+      (_crossDeviceLinkMethods && _crossDeviceLinkMethods.length < 1)
+    ) {
+      return defaultViewRendersMap
+    }
+    if (
+      _crossDeviceLinkMethods &&
+      configHasInvalidViewIds(_crossDeviceLinkMethods)
+    ) {
+      return defaultViewRendersMap
+    }
```
rabelloo, line 348: 💅🏻 [nitpick]

Could simplified to 2 expressions (or 3 if you count the optional chaining operator) if we take advantage of the fact an undefined array and an empty one are essentially the same fail condition.

```suggestion
    if (
      !_crossDeviceLinkMethods?.length ||
      configHasInvalidViewIds(_crossDeviceLinkMethods)
    ) {
      return defaultViewRendersMap
    }
```

src/components/crossDevice/CrossDeviceLink/index.js
```diff
@@ -307,18 +327,48 @@ class CrossDeviceLinkUI extends Component {
...
-    const currentViewRender = {
-      qr_code: this.renderQrCodeSection,
-      sms: this.renderSmsLinkSection,
-      copy_link: this.renderCopyLinkSection,
-    }[currentViewId]
+    const requiredViewRenders = this.getRequiredViewRenders()
+    const currentViewRender = requiredViewRenders[currentViewId]
+    const visibleViewOptions = SECURE_LINK_VIEWS.filter(
+      (view) => (view = Object.keys(requiredViewRenders).includes(view.id))
+    )
```
rabelloo, line 371: 💅🏻 [nitpick]

I don't think `view =` is doing anything, is it?
Also, we can use keyword `in` to check whether the object has that key defined and reduce time complexity.

```suggestion
    const visibleViewOptions = SECURE_LINK_VIEWS.filter(
      (view) => view.id in requiredViewRenders
    )
```

rabelloo (approved):
Approach looks good, some nitpicks and one question.

src/components/crossDevice/CrossDeviceLink/index.js
kopijunkie: Keep forgetting that Set exists. 😅 Didn't realise `Set.has()` is faster than `Array.includes()` though....TIL

src/components/crossDevice/CrossDeviceLink/index.js
kopijunkie: You're right. I originally tried to use `console.warn` and even `console.error` but they were not showing up on the browser console. Tried again today & turns out it was because of the console's text filter having some text. 🤦‍♀️  

Am avoiding throwing an error since we don't want to break the UX for the end user if the integrator chooses to still ignore the warning during integration.

src/components/crossDevice/CrossDeviceLink/index.js
kopijunkie: Interesting. I've never used that approach with the `in` keyword for that check. And yes, `view =` isn't required there. Might have got left there following a rebase and/or me cleaning up the code.

src/components/crossDevice/CrossDeviceLink/index.js
rabelloo: Yea Sets and Maps (and plain objects to a degree) have O(1) access, but Arrays are O(n) because a lookup is needed.
Arrays have O(1) access by index, but the index is just the position, not an identifier for the element.

i.e. it's like checking `array[index] === foo` versus `for (const element of array) element === foo`

src/components/crossDevice/CrossDeviceLink/index.js
rabelloo: It's almost the same thing as using a Set too, like `new Set(Object.keys(requiredViewRenders))`.
Except plain objects already are indexed by their keys so we don't need the Set.

src/components/crossDevice/CrossDeviceLink/index.js
```diff
@@ -340,32 +384,34 @@ class CrossDeviceLinkUI extends Component {
...
+                  )}
+                  ref={(node) => (this.viewOptionBtn = node)}
+                  onClick={preventDefaultOnClick(() =>
+                    this.handleViewOptionSelect(view.id)
+                  )}
+                  key={`view_${view.id}`}
+                >
+                  {translate(view.label)}
+                </a>
+              ))}
```
ethan605, line 414: I'm curious what'll be rendered if 2 options were provided (1 visible, 1 other). Does that break the view or display as normal?

ethan605 (commented):
LGTM! Only a minor question

src/components/crossDevice/CrossDeviceLink/index.js
kopijunkie: The view will still display as normal, but end user will only see the 1 other option under the `-- or choose an alternative method --` divider.

ethan605 (approved):
LGTM!

stefaniacardenas (approved):

PR #1534, opened 2021-08-17
Improvement/cx 5496 x device scam warning

kopijunkie:
# Feature

New intro screen to be displayed when the end user starts on the Cross Device IDV flow on their mobile devices.
Screenshot 2021-08-17 at 11 04 57

## Checklist

- [x] Have any new strings been translated or the existing ones changed?


src/components/crossDevice/ClientSessionLinked/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+import PageTitle from '../../PageTitle'
+import { trackComponent } from '../../../Tracker'
+import { useLocales } from '~locales'
+import theme from '../../Theme/style.scss'
+import style from './style.scss'
+
+import type { StepComponentBaseProps } from '~types/routers'
+
+type Props = {
+  nextStep: () => void
```
ethan605, line 15: Maybe I was wrong to add this redundant `nextStep` somewhere else (like [this](https://github.com/onfido/onfido-sdk-ui/blob/9b2d8eb225770e55822b05cd31c97c65ef2985aa/src/components/FaceVideo/Intro.tsx#L19-L21)), but I believe you can use `StepComponentBaseProps` directly because `nextStep` was already in that type: https://github.com/onfido/onfido-sdk-ui/blob/9b2d8eb225770e55822b05cd31c97c65ef2985aa/src/types/routers.ts#L95

ethan605 (approved):
LGTM, just a minor typing suggestion

src/components/crossDevice/ClientSessionLinked/index.tsx
kopijunkie: Ah, I see. I happened to see it done that way in a random TS component file I opened so just went with it. Will try with `StepComponentBaseProps` directly instead.

kopijunkie: Looks like it isn't possible. There is a [TS error](https://github.com/onfido/onfido-sdk-ui/pull/1534/checks?check_run_id=3370992615#step:5:184) that `Object is possible 'undefined'` when I do this. 🤔

ethan605: Humm, kinda weird. Anw we can try to get rid of this with another PR.

src/components/Router/StepComponentMap.ts
```diff
@@ -171,17 +176,24 @@ const buildFaceComponents = (
...
   shouldUseCamera?: boolean,
-  mobileFlow?: boolean
+  mobileFlow?: boolean,
+  showClientSessionLinkedIntro?: boolean
 ): ComponentType<StepComponentProps>[] => {
-  // @TODO: convert FaceVideoIntro, FaceVideoCapture, FaceVideoConfirm to TS
-  const allVideoSteps = [FaceVideoIntro, FaceVideoCapture, FaceVideoConfirm]
+  // @TODO: convert FaceVideoCapture, FaceVideoConfirm to TS
+  const allVideoSteps = showClientSessionLinkedIntro
+    ? [FaceVideoIntro, ClientSessionLinked, FaceVideoCapture, FaceVideoConfirm]
```
stefaniacardenas, line 195: I would challenge the decision of displaying this screen after the Selfie/Video Intro screen. I think it should be the first screen that a user sees when they land on the `mobileFlow`. I think that a user might be inclined to skip the information on this screen if they are already getting ready for the capture.

src/components/crossDevice/ClientSessionLinked/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+} & StepComponentBaseProps
+
+const ClientSessionLinked: FunctionComponent<Props> = ({ nextStep }) => {
```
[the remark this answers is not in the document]
stefaniacardenas: By looking at the `StepComponentMap` I figured out that this component will not be shown as soon as the user opens the cross-device tab. As mentioned in another comment, I think this might be quite confusing for the user and looks like it's adding a bit of complexity on our side too.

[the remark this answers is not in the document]
kopijunkie: > I would rename this component to `CrossDeviceClientIntro` because it's always shown as soon as you open the cross-device tab, right?
Totally! I was just thinking when working on the follow on ticket to this that the component needed a rename to something like what you've suggested to make it clearer what it is.

src/components/Router/StepComponentMap.ts
```diff
@@ -294,22 +322,45 @@ const buildDocumentComponents = (
...
+
+const buildPoaComponents = (
+  showClientSessionLinkedIntro: boolean | undefined
+): ComponentType<StepComponentProps>[] => {
+  // @TODO: convert PoAIntro, SelectPoADocument, PoAGuidance, PoACapture, DocumentFrontConfirm to TS
+  const preCaptureComponents = [PoAIntro, SelectPoADocument, PoAGuidance]
+  const captureComponents = [PoACapture, DocumentFrontConfirm]
+  // @ts-ignore
+  return showClientSessionLinkedIntro
+    ? [...preCaptureComponents, ClientSessionLinked, ...captureComponents]
```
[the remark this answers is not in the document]
kopijunkie: I thought so also, but I found that the Cross Device flow's step index becomes out of sync by 1 without this. The same applies for the Document, Face steps. 
Though I could readjust the index in the [`MainRouter`'s `onFlowChange` function](https://github.com/onfido/onfido-sdk-ui/blob/35feb4e8c55864f7a49f6dfbbdc395eb5faeb138/src/components/Router/MainRouter.tsx#L84). 🤔 
This should also then address the concern in your earlier comment about when this new screen is actually displayed to the user from looking at `StepComponentMap`. We can chat further on Slack if needed.

src/components/crossDevice/ClientSessionLinked/index.tsx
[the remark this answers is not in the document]
kopijunkie: > By looking at the `StepComponentMap` I figured out that this component will not be shown as soon as the user opens the cross-device tab. As mentioned in another comment, I think this might be easily missed by the user and looks like it's adding a bit of complexity on our side too.

It actually is. Or rather, it is the first screen the end user sees on loading the Cross Device link on their device browser and the user will need to tap on the "Continue" button on the screen to continue their cross device capture flow...unless I've misunderstood what you meant here.

src/components/Router/StepComponentMap.ts
kopijunkie: I can rename this to make it clearer, but the screen is only displayed when they first land on the `mobileFlow` Cross Device client. Should be clearer also with the other proposed change to address your earlier feedback. 
Was trying to avoid complexity or even having to make the change here in general but this seemed like the best/least complex place to implement it. Agree that this does add further to the complexity here. Open to suggestions if you can see a better way to add this screen into the flow.

stefaniacardenas: I see, thanks for the clarification. In that case, there is probably some logic I'm not familiar with in the Router that handles it. Might be good to add a comment or refactor because it's not obvious from the stepComponentMap that ClientSessionLinked is the first screen a user would see on the cross device client.

src/components/Router/StepComponentMap.ts
```diff
@@ -120,39 +124,46 @@ const buildCaptureStepComponents = (
...
+  const showCrossDeviceClientIntroForDocumentStep =
+    mobileFlow && firstCaptureStepType === 'document'
+  const showCrossDeviceClientIntroForPoaStep =
+    mobileFlow && firstCaptureStepType === 'poa'
 
   return {
     welcome: [Welcome],
     userConsent: [UserConsent],
-    face: buildFaceComponents(faceStep, deviceHasCameraSupport, mobileFlow),
+    face: buildFaceComponents(
```
stefaniacardenas, line 141: Maybe something like this would simplify the logic in this component?
```
face: [...crossDeviceClientIntro, ...buildFaceComponents(faceStep, deviceHasCameraSupport, mobileFlow)],
```

And crossDeviceClientIntro would be
```
const crossDeviceClientIntro = mobileFlow ? [CrossDeviceIntro] : []

Comment (kopijunkie) [reply src/components/Router/StepComponentMap.ts:141]: It looks like it's not really required after my refactor, but will definitely keep this in mind. :)

File: src/locales/it_IT/it_IT.json [hunk 6]
```diff
@@ -46,6 +46,14 @@
         "subtitle": "Ora puoi tornare al tuo computer per continuare",
         "title": "Caricamenti riusciti"
     },
+    "cross_device_session_linked": {
+        "button_primary": "Continua",
+        "info": "Doppio controllo",
```
stefaniacardenas, line 51: This does not really make sense in Italian. It should be "Assicurati di"

src/components/Router/StepComponentMap.ts
```diff
@@ -120,30 +124,54 @@ const buildCaptureStepComponents = (
...
+  const showCrossDeviceClientIntroForDocumentStep =
+    mobileFlow && firstCaptureStepType === 'document'
+  const CrossDeviceClientIntroForDocumentStep = showCrossDeviceClientIntroForFaceStep
+    ? [CrossDeviceClientIntro]
+    : []
+  const showCrossDeviceClientIntroForPoaStep =
+    mobileFlow && firstCaptureStepType === 'poa'
+  const CrossDeviceClientIntroForPoaStep = showCrossDeviceClientIntroForFaceStep
+    ? [CrossDeviceClientIntro]
+    : []
```
stefaniacardenas, line 145: I'm starting to think that the `crossDeviceClientIntro` should not be in the component array for any of the steps. I think you should display it conditionally if it's `mobileFlow` and if it has not been shown yet. You could use a flag in the redux store for that and use it in the `CrossDeviceMobileRouter` component.

src/locales/it_IT/it_IT.json
kopijunkie: @raemarielee FYI ^

src/components/Router/StepComponentMap.ts
kopijunkie: Wouldn't that still need the screen to be added in in the `StepComponentMap`? Or were you thinking something like adding it in as an additional step type in the `CrossDeviceMobileRouter` base on what's state the flag is in the redux store?

stefaniacardenas: > Or were you thinking something like adding it in as an additional step type in the CrossDeviceMobileRouter base on what's state the flag is in the redux store

Yes, you might not even need to use redux and it could be handled by a state in `CrossDeviceMobileRouter`

stefaniacardenas (approved):
This is much clearer! Thanks for the change!

Issue #1539, opened 2021-08-24
Component is auto-focusing after loading


lf94:
I'm going to open a PR for this; this issue is to quickly record that there is a problem.

```
    onfido-sdk-ui/src/components/PageTitle/index.tsx:29.1,29.57
    containerRef.current && containerRef.current.focus()
```
The PageTitle component is causing the page to scroll to it after mounting. This is problematic as the component is loaded with other information the user needs to read and fill out before reaching the component.

Ludovic71:
Same issue here, any change that this issue is being fixed soon ? If there is some accessibility concerns, one solution could be to pass a props to decide to auto-focus on the title or not. Thanks!

lf94:
@Ludovic71 I don't work for Onfido but I'll probably do this today, thank you for reminding me. :slightly_smiling_face:

lf94:
I notice that `PageTitle` is everywhere, and would involve a bunch of changes. I'm not sure if Onfido would prefer to pass a "focused" property all over, or to use some sort of global state to pass it around.

Anyone?

Ludovic71:
Thanks @lf94 , sorry I indeed though you work for Onfido. 

For this question about Welcome Panel, we don't use it for example and directly start at the document Panel. Also does a focus on a span title really makes sense, instead of actionable element like button/input ? 

But I agree that it would be a breaking change for people relying on this focus and passing a props to decide weather need to focus or not (enabled by default to keep backward compatibility) seems the best choice ?

kopijunkie:
Hi @lf94 

Thank you for your PR contribution to address this issue. 

We did indeed implemented the focus on PageTitle for accessibility reasons to have screen reader focus start at the PageTitle when a user is stepping through each screen on the SDK flow. 

Just to be sure that we understand the issue correctly, is this UX causing your site/app to scroll to where this element is on initialising the SDK but this is not desirable for in yours & @Ludovic71 case as there is content on your site/app that you'd like the user to see first?

Ludovic71:
@kopijunkie , yes for our case we have TextBoxes to fill in User Name and LastName in a form. The auto focus should remain on the first TextBox or the TextBox users are typings, but because of this pageTitle.focus() there is an annoying behavior that users can't type in the textBox anymore until re-select it. 

FYI, we dont use the welcome Panel and  1st panel to show is current document but can be configurable or different based on some user  data. So it's important that we decide to not have onfido step-wizard auto-focus when it's loaded

[...]

PR #1546, opened 2021-09-01
Add option to focus on Welcome title

lf94:
# Problem

https://github.com/onfido/onfido-sdk-ui/issues/1539

# Solution

Add a `focused` property to the Welcome step.

## Checklist

### DIFF
--- a/src/components/PageTitle/index.tsx
+++ b/src/components/PageTitle/index.tsx
@@ -12,6 +12,7 @@ type Props = {
   smaller?: boolean
   subTitle?: string
   title: string
+  focused?: boolean
 }
 
 const PageTitle: FunctionComponent<Props> = ({
@@ -19,6 +20,7 @@ const PageTitle: FunctionComponent<Props> = ({
   subTitle,
   smaller,
   className,
+  focused
 }) => {
   const isFullScreen = useSelector<RootState, boolean | undefined>(
     (state) => state.globals.isFullScreen
@@ -26,7 +28,7 @@ const PageTitle: FunctionComponent<Props> = ({
   const containerRef = useRef<HTMLSpanElement>(null)
 
   useEffect(() => {
-    containerRef.current && containerRef.current.focus()
+    containerRef.current && focused && containerRef.current.focus()
   }, [title, subTitle])
 
   return (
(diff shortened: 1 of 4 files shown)


lf94:
The default focusing behavior is preserved. Developer must opt-out of focusing on Welcome title.

src/components/PageTitle/index.tsx
```diff
@@ -12,21 +12,23 @@ type Props = {
   smaller?: boolean
   subTitle?: string
   title: string
+  focused?: boolean
```
Ludovic71, line 15: maybe rename "autoFocus" instead

src/components/PageTitle/index.tsx
```diff
@@ -12,21 +12,23 @@ type Props = {
...
+  focused
 }) => {
   const isFullScreen = useSelector<RootState, boolean | undefined>(
     (state) => state.globals.isFullScreen
   )
   const containerRef = useRef<HTMLSpanElement>(null)
 
   useEffect(() => {
-    containerRef.current && containerRef.current.focus()
+    containerRef.current && focused && containerRef.current.focus()
```
Ludovic71, line 31: Can we pick this value from props to a provider so PageTitle avoid to autofocus on other views too instead of only welcome panel ?

lf94: I'm not sure it's good behavior to be "all in" or not. Some Onfido dev will have to make the say. I agree with you.

Ludovic71: Yes, I agree that some onfido dev should evaluate this and eventually add the necessary tests with this change. Hope this PR can highlight the issue at least, thks for making it  !

kopijunkie: We've actually had a need to have the focus not pulled back to the PageTitle in a/some screens/scenarios. See PR #1560 where it is passed down as a prop.

But we have a bit more control internally within our code to choose which screen gets to pass the prop down. We'll need to have a think about how to best make this available as an integrator controllable option since as your discussions in the original issue raised show that integrators don't always start from the Welcome screen and order of steps also differ and an "all in" approach is not ideal IMO.

stefaniacardenas:
Closing this PR as the focus behaviour was changed by introducing the `autoFocusOnInitialScreenTitle` option in version `6.16.0`

Issue #1519, opened 2021-07-26
Select Issuing Country Input Not Displayed


MMA15:
So the current issue I am running into is the country text is displayed but the color will not change. 
This happens, on the Select Country screen, after selecting a country. The default color of white, prevents it from being seen. I tried setting `colorContentBody` to black but still no luck. Changing the `colorContentBody` changed the colors for all the content to black as it should, but it did not for the country input, this remained white.

This is for browser using React. Onfido-sdk-ui": "^6.11.1

Posted images of it below.

<img width="565" alt="Screen Shot 2021-07-26 at 10 33 39 AM" src="https://user-images.githubusercontent.com/44588455/127007548-eda32a86-0828-461a-b435-a976480429d7.png">
<img width="550" alt="Screen Shot 2021-07-26 at 10 33 54 AM" src="https://user-images.githubusercontent.com/44588455/127007556-2b772792-ea0e-4091-bc19-625a5598b308.png">
<img width="487" alt="Screen Shot 2021-07-26 at 10 34 02 AM" src="https://user-images.githubusercontent.com/44588455/127007559-3a5cc28c-1683-494e-b59a-6b1db42c2f7f.png">

customUI: { 
  colorContentBody: "#000000",
}

kopijunkie:
Hi @MMA15 

Which browser(s) are you seeing this behaviour on?

MMA15:
Hi @kopijunkie this is on Google Chrome as well Mozilla Firefox.

kopijunkie:
Would you be able to inspect the input element and share the screenshot of that so that I can see what styles are being applied to it? In our experience sometimes external styling might be overriding the SDK's styles depending on the specificity level but just wanted to be sure before making any assumptions or suggestions for fixing the issue.

MMA15:
<img width="1617" alt="Screen Shot 2021-07-26 at 1 10 33 PM" src="https://user-images.githubusercontent.com/44588455/127030530-3aad8005-14c8-4b74-8a5b-98b5f8a16185.png">
@kopijunkie screenshot of the input element and its css. 

I originally did think it was being overridden by one of our styles and tried to change the css of our inputs to see if that would change it but it remained unaffected.

MMA15:
https://eternal-frontend-dev-pr-111.onrender.com/settings This is a test environment we set up for the integration. 

`test@test.gg : test@test.gg` you can login with this, if my screenshot isn't clear enough.

[...]

PR #1548, opened 2021-09-06
Bugfix/Country Selector dropdown issues

kopijunkie:
# Problem

1. In Country Selector screen, the country search's suggestions dropdown list closes on clicking on the scrollbar. This means the dropdown is only scrollable using a mouse's middle scroll button. Issue does not happen on mobile browsers.
2. If the integrator's host app/site has a default text colour set at `input` element level this will get applied to the country search's text input also instead of the custom SDK body text colour, if set. #1519 

# Solution

1. Intercept mouse down event if user clicks on the dropdown list menu area where the scrollbar is as the `accessible-autocomplete/preact` library is picking it up as a selection confirm click event.
2. Set a text colour for `.custom__input` mapped to `--osdk-color-content-body` token variable.
3. Extra: Refactored component to TS

## Checklist


src/components/CountrySelector/index.tsx
```diff
@@ -78,13 +88,28 @@ class CountrySelection extends Component {
...
+    ) {
+      event.preventDefault()
+    }
+  }
+
   componentDidMount() {
     if (this.props.idDocumentIssuingCountry) {
       this.props.actions.resetIdDocumentIssuingCountry()
     }
+    document.addEventListener('mousedown', this.handleMenuMouseClick)
```
rabelloo, line 109: It would be good to remove this event listener on `componentWillUnmount`

rabelloo (approved):

src/components/CountrySelector/index.tsx
kopijunkie: Good point. Forgot to do this clean up.

ethan605 (approved):
LGTM! It should be great if you can provide a minimum unit test for the newly converted TS component 🙏

PR #1554, opened 2021-09-09
Bugfix/cx 5266 handle live_videos validation errors

kopijunkie:
# Problem

When a user takes a video that is very short (and by very short I mean you literally have to click through the video buttons really really fast) and submit it, the Onfido API returns a validation error:

```
{"attachment_file_size":["File size must be larger than zero"],"attachment":["File size must be larger than zero"]}
```

The SDK does not have the `attachment_file_size` error in the list of validation errors, therefore the user might not see any error (they just see the spinner) or they see the generic error message that implies that the connection was lost.

This error or any errors from the `video` element in the Confirm screen's previewer should be handled by the SDK and displayed to the user so that they can be guided to retake the video. 

# Solution

For all unhandled API errors or unexpected SDK UI errors, user should see the generic "please try again" message (i.e currently - connection lost)
- the generic "connection lost, please try again" message should be updated to just say "something's gone wrong, please try again"

## Checklist

### DIFF
--- a/src/components/CaptureViewer/CaptureVideoViewer.tsx
+++ b/src/components/CaptureViewer/CaptureVideoViewer.tsx
@@ -1,24 +1,33 @@
 import { h, FunctionComponent } from 'preact'
 import classNames from 'classnames'
 import { withBlobPreviewUrl } from './hocs'
+import { trackException } from '../../Tracker'
 import style from './style.scss'
 
 type CaptureVideoViewerProps = {
   ariaLabel: string
   className?: string
   previewUrl?: string
+  onVideoError: () => void
 }
 
 const CaptureVideoViewer: FunctionComponent<CaptureVideoViewerProps> = ({
   ariaLabel,
   className,
   previewUrl,
+  onVideoError,
 }) => (
   <div className={classNames(style.videoWrapper, className)}>
     <video
       aria-label={ariaLabel}
       className={style.video}
       src={previewUrl}
+      onError={(event) => {
+        const target = event.target as HTMLVideoElement
+        const error = target.error as MediaError
+        trackException(`${error.code} - ${error.message}`)
+        onVideoError()
+      }}
       controls
     />
   </div>
--- a/src/components/Confirm/Confirm.js
+++ b/src/components/Confirm/Confirm.js
@@ -30,6 +30,7 @@ const CALLBACK_TYPES = {
   video: 'onSubmitVideo',
   document: 'onSubmitDocument',
 }
+const UNEXPECTED_ERROR = 'REQUEST_ERROR'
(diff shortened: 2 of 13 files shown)


stefaniacardenas (approved):
LGTM!

src/components/Confirm/Confirm.js
```diff
@@ -30,6 +30,7 @@ const CALLBACK_TYPES = {
   video: 'onSubmitVideo',
   document: 'onSubmitDocument',
 }
+const UNEXPECTED_ERROR = 'REQUEST_ERROR'
```
ethan605, line 33: Minor: extra line break should be good here

```suggestion
const UNEXPECTED_ERROR = 'REQUEST_ERROR'

```

src/components/Confirm/Previews.js
```diff
@@ -1,6 +1,6 @@
 import { h } from 'preact'
 import classNames from 'classnames'
-import { localised } from '../../locales'
+import { useLocales } from '../../locales'
```
ethan605, line 3: I would say this is a good conversion, but seems a bit unrelated to the whole PR

src/components/CaptureViewer/CaptureVideoViewer.tsx
```diff
@@ -1,24 +1,33 @@
...
   className,
   previewUrl,
+  onVideoError,
 }) => (
   <div className={classNames(style.videoWrapper, className)}>
     <video
       aria-label={ariaLabel}
       className={style.video}
       src={previewUrl}
+      onError={(event) => {
```
ethan605, line 25: What's the default type of `event` here? Can we explicitly infer its type here to avoid (extra) casting inside the function?

ethan605 (changes requested):
A minor suggestion and an open question

src/components/Confirm/Previews.js
kopijunkie: Ah, this is my bad. I was initially going to try to do a TS conversion but decided against it & left this behind. Will revert it to have only relevant changes in this PR.

src/components/CaptureViewer/CaptureVideoViewer.tsx
kopijunkie: In React I think it would have been `SyntheticEvent<HTMLVideoElement>` but Preact does not have Synthetic events so I didn't know of any other way to handle the `property not existing on EventTarget` TS error that was originally appearing without the extra casting.

ethan605: Okay fair enough!

ethan605 (approved):

PR #1556, opened 2021-09-09
Release/6.14.0

kopijunkie:
### Added

- UI: Added new Intro screen when user begins the Cross Device flow on their mobile device.
- Public: Added SDK configuration options for integrators to customize product name, copy and/or logo image for the new Cross Device Mobile Client Intro screen.

### Fixed

- Public: Fixed custom SDK body text color set in `customUI.colorContentBody` option not getting applied to Country Selector text input if an element level text colour has been set in host app/site's stylesheet.
- UI: Fixed Country Selector dropdown menu closing on clicking on scrollbar.


.github/workflows/publish_release.yml
```diff
@@ -70,6 +73,19 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
 
+      - name: Extract release notes from CHANGELOG
+        if: "!contains(github.ref, '-rc.')"
```
stefaniacardenas, line 77: You only want to get the changelog if it's a full release, right? In that case, you can do

```suggestion
        if [ "$VERSION" == "$LATEST_TAG" ]
```

.github/workflows/publish_release.yml
```diff
@@ -70,6 +73,19 @@ jobs:
...
+        id: extract-release-notes
+        uses: ffurrer2/extract-release-notes@v1
+
+      - name: Create Github Release
+        if: "!contains(github.ref, '-rc.')"
```
stefaniacardenas, line 82: Same here

MIGRATION.md
```diff
@@ -2,6 +2,19 @@
 
 The guides below are provided to ease the transition of existing applications using the Onfido SDK from one version to another that introduces breaking API changes.
 
+## `6.12.0` -> `6.14.0`
```
stefaniacardenas, line 5: Shouldn't this be `6.13.0`?
```suggestion
## `6.13.0` -> `6.14.0`
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "onfido-sdk-ui",
-  "version": "6.13.0",
+  "version": "6.14.0-rc.8",
```
stefaniacardenas, line 3: Why 8? 😅 Were you testing the new action?

ethan605 (approved):
LGTM!

.github/workflows/publish_release.yml
kopijunkie: Yes. I thought I did test this condition & it didn't work, but there doesn't seem to be a commit for it in my history. I might have only done it locally. Can try running another RC with this just to be sure.

MIGRATION.md
kopijunkie: I'm actually not sure what the convention is, but was following on from the last version listed in the doc.

package.json
kopijunkie: Yeah, not found a way to be able to test locally first. That's why it got bumped until 8. 😅

MIGRATION.md
stefaniacardenas: I think it's supposed to list the difference between the current version and the previous

stefaniacardenas (approved):

PR #1560, opened 2021-09-16
Improvement/return focus to enlarge image button on preview close

kopijunkie:
# Problem
On Document capture confirmation screen, when image preview is enlarged and screen reader user clicks on “Close” toggle button, they return to the confirmation screen but there’s no element in focus. The user has to go through the top of the page to search for an element to focus on again. 

Additionally in latest SDK designs the "Enlarge image" toggle button is at the bottom of the preview frame.

# Solution
- Make sure that when the enlarged image preview is closed, the focus is returned to the "Enlarge image" toggle button 
- Move "Enlarge image" toggle to bottom of preview frame

## Checklist


src/components/Confirm/Previews.js
```diff
@@ -124,7 +124,12 @@ const Previews = localised(
               }}
             />
           ) : (
-            <PageTitle title={title} smaller={true} className={style.title} />
+            <PageTitle
+              title={title}
+              smaller={true}
```
ethan605, line 129: Should this be
```suggestion
              smaller
```
only

src/components/PageTitle/index.tsx
```diff
@@ -12,22 +12,24 @@ type Props = {
...
   const isFullScreen = useSelector<RootState, boolean | undefined>(
     (state) => state.globals.isFullScreen
   )
   const containerRef = useRef<HTMLSpanElement>(null)
 
   useEffect(() => {
-    containerRef.current && containerRef.current.focus()
-  }, [title, subTitle])
+    containerRef.current && shouldAutoFocus && containerRef.current.focus()
+  }, [title, subTitle, shouldAutoFocus, isFullScreen])
```
ethan605, line 32: I'd love to have an extra comment here to explain why `isFullScreen` needs to be in the dependencies list

ethan605 (approved):

src/components/Confirm/Previews.js
```diff
@@ -124,7 +124,12 @@ const Previews = localised(
...
+              className={style.title}
+              shouldAutoFocus={methodNamespace !== 'doc_confirmation'}
```
stefaniacardenas, line 131: Does this mean that for documents the confirmation screen won't be announced?

stefaniacardenas (changes requested):

src/components/PageTitle/index.tsx
kopijunkie: Ah, that's my mistake. I was initially trying to use `isFullScreen` to determine if it should be auto focused on but that wasn't the right conditional to use. Will remove it from the dependencies list.

src/components/Confirm/Previews.js
kopijunkie: This only affects the focused element to not be the screen (page) title like for this case where on closing the preview we want the focus to start from the `Enlarge image` button. Title is always announced as the component's `aria-live` is set to "assertive"

stefaniacardenas (approved):

ethan605 (approved):

PR #1572, opened 2021-09-29
Sending Analytics events to new Onfido endpoint

stefaniacardenas:
# Problem
Currently, we use Woopra to track analytics events. We want to move away from using Woopra and use an in-house analytics service.

# Solution
Send data to the new Onfido analytics endpoint `v3/analytics`. The events will be sent to both Woopra and to the new service until we are ready to switch Woopra off.

## Checklist


bjacquet (approved):
Calls to `Tracker.sendEvent()` require the second argument to be a string of either `screen`,  `action`, or `flow`. I don't like having literal strings appearing out of the blue, so instead of having:
```javascript
Tracker.sendEvent('completed flow', 'flow')
```
I prefer to see something like this instead:
```javascript
Tracker.sendEvent('completed flow', Tracker.TrackedEventTypes.flow)
```
but better still would be:
```javascript
# src/Tracker/index.ts
const sendFlowEvent = (
  eventName: TrackedEventNames,
  properties?: Record<string, unknown>
): void => sendEvent(eventName, 'flow', properties?)

# src/components/App/ModalApp.tsx
Tracker.sendFlowEvent('completed flow')
```

What do you think of this?

ethan605 (approved):
Just some minor suggestions. Otherwise LGTM 🙏

PR #1590, opened 2021-10-26
Bugfix/Microsoft Surface tablet layout issues

kopijunkie:
# Problem

On Microsoft Surface tablets there are 2 layout issues: 

1. In portrait mode, the camera view during selfie capture step is shifted too far to the left so looks like there is a black block covering about half of the view. Camera view is fine when in landscape mode.
2. On some integrations, the Country Selector's search text input box height is stretched to be the full container height so user cannot see the primary button to proceed to next step without scrolling.

# Solution

1. Remove redundant left margin value and additional  `.adjustForReducedResolution` class, styling
2. Set `height` for the Country Selector search text input to `auto` instead of `100%`

## Checklist


src/components/Camera/style.scss
```diff
@@ -102,8 +102,9 @@
 .webcamContainer {
   position: absolute;
   top: 0;
+  left: 0;
```
stefaniacardenas, line 105: I think you need to change this on other media queries as well

kopijunkie: Good point. The layout displays fine on the mobile devices I tested as is so didn't think to change it, but I'll check if they are required to be sure.

stefaniacardenas (approved):

PR #1602, opened 2021-11-03
Improvement/cx 6912 upgrade react phone number input

kopijunkie:
# Task
Web SDK is still on v2 of `react-phone-number-input` and there are some NPM vulnerabilities, React version conflicts that require upgrading to v3.

## Checklist


josokinas (approved):

kopijunkie:
> I can see that `libphonenumber-js` is back on the `onfido.min.js` bundle. It should only be in crossDevice bundle to prevent unnecessary size increase of the main bundle. You can compare the [report](https://1602-pr-onfido-sdk-ui-onfido.surge.sh/reports/bundle_idv_dist_size.html) from this change with the [report for latest](https://latest-onfido-sdk-ui-onfido.surge.sh/reports/bundle_idv_dist_size.html).
> 
> The screenshot below shows the duplication of `libphonenumber-js` in red.
> 
> Screenshot 2021-11-09 at 12 54 07

Unfortunately it looks like this is a side effect of upgrading to v3 of this library. The architecture has been changed in a way that the `getCountryCodes(labels)` method we were using has been removed in favour of this `getCountries()` method. Additionally, contrary to their [CHANGELOG](https://gitlab.com/catamphetamine/react-phone-number-input/-/blob/master/CHANGELOG.md), `getCountries()` actually requires a `metadata` object which only exists in `libphonenumber-js`. 

We should have a chat when you're back about the impact of this change or pair to see if there is a better workaround I've not tried yet.

src/index.tsx
```diff
@@ -82,7 +82,7 @@ const experimentalFeatureWarnings = ({ steps }: NormalisedSdkOptions) => {
 const isSMSCountryCodeValid = (smsNumberCountryCode: string) => {
   // If you need to refactor this code, remember not to introduce large libraries such as
   // libphonenumber-js in the main bundle!
-  const countries = getCountryCodes(labels)
+  const countries = getCountries(metadata) as string[] // react-phone-number-input returns an array of CountryCode type contrary to documentation
```
stefaniacardenas, line 85: Is this the only way of doing this validation? I don't like the idea of importing this large library globally for this validation only tbh...

kopijunkie: Agreed about having to introduce the large library just for the country code validation. I've had a think & what I can try is hacky, but we could make use of the `labels` value to do the validation like it used to. The file it was originally imported from still exists, just renamed/restructured.

package.json
```diff
@@ -93,15 +93,15 @@
     "files": [
       {
         "path": "./dist/onfido.min.js",
-        "maxSize": "240 kB"
+        "maxSize": "260 kB"
```
stefaniacardenas, line 96: Do we still need to bump this?

kopijunkie: Good point. I'll revert this change.

stefaniacardenas (approved):

PR #1618, opened 2021-11-12
ignore the release folder for eslint

it-ony:
# Problem

eslint brings up a lot of warnings in the release directory as the scripts use nodejs require. The scripts itself work, but they're not from the correct style as configured in eslint.

# Solution

added the release directory to .eslintignore.

## Checklist


.eslintignore
```diff
@@ -2,3 +2,4 @@ test/*.conf.js
 node_modules
 test/mock-server/**/*.js
 coverage
+release/
```
kopijunkie, line 5: IMO it would be better to use the `eslint-disable`/`eslint-enable` toggle for the `@typescript-eslint/no-var-requires rule on [release/release.js](https://github.com/onfido/onfido-sdk-ui/blob/master/release/release.js), [release/releaseConfig.js](https://github.com/onfido/onfido-sdk-ui/blob/master/release/releaseConfig.js) files where appropriate rather than exclude the entire file from linting since that is the only warning(s).

kopijunkie (changes requested):

.eslintignore
it-ony: If you want to go for it, then please follow this approach. Having 40 something warnings from the linter just leads to the fact that nobody is giving a look at all and basically linting can be skipped if it isn't enforced. 

Linting should be done without any warning.

kopijunkie: True. I didn't think there were that many warnings from this folder, but if that's the case then I'd agree to ignore the folder for now.

kopijunkie (approved):

PR #1620, opened 2021-11-12
enable sentry again as CX-6470 tells that sentry is fixed

it-ony:
# Problem

FIXME in the code

```
FIXME: Temporarily disabled to allow release build to pass. To be fixed in CX-6470
```

Ticket states that the problem seems to be fixed

# Solution

Enable it again

## Checklist


stefaniacardenas (approved):
🙏

stefaniacardenas:
I'm not sure that this action will succeed. I think we will have to do a test release to make sure this is working, so that the action does not fail when we are in the middle of a real release. I've added more details in CX-6470. I will merge this one into a separate PR. The acceptor will have to create a test release to make sure this is working.

kopijunkie:
> I'm not sure that this action will succeed. I think we will have to do a test release to make sure this is working, so that the action does not fail when we are in the middle of a real release. I've added more details in CX-6470. I will merge this one into a separate PR. The acceptor will have to create a test release to make sure this is working.

Agreed. From what I remember of this particular job, during the release it fails as the `sentry-cli` command doesn't exist despite attempting to install it with `curl -sL https://sentry.io/get-cli/ | bash` command first. But this can be re-verified by acceptor with a test release.

kopijunkie (approved):

PR #1641, opened 2021-12-03
Chore/update integration tests expired token expected response

kopijunkie:
# Problem

API endpoints for uploading `documents`/`live_photos`/`live_videos` has changed to return no info in the error response 
on attempting to upload a file with an expired JWT token. This causes the integration tests for expired token scenario to fail and hence CI build also fails.

# Temporary Solution

Update expected error response for document, selfie, video integration tests to unblock builds while API is fixed to return the intended (new) error response.

## Checklist

### DIFF
--- a/src/components/utils/__integrations__/onfidoApi/selfie.integration.js
+++ b/src/components/utils/__integrations__/onfidoApi/selfie.integration.js
@@ -13,8 +13,8 @@ import {
 import { API_URL, PATH_TO_RESOURCE_FILES } from '../helpers/testUrls'
 import {
   EXPIRED_JWT_TOKEN,
+  ORIGINAL_EXPECTED_EXPIRED_TOKEN_ERROR,
   EXPECTED_EXPIRED_TOKEN_ERROR,
-  TEMP_EMPTY_ERROR,
 } from '../helpers/mockExpiredJwtAndResponse'
 
 let jwtToken = null
@@ -68,7 +68,7 @@ describe('API uploadFacePhoto endpoint', () => {
     expect.hasAssertions()
     const onErrorCallback = (error) => {
       try {
-        expect(error).toEqual(TEMP_EMPTY_ERROR)
+        expect(error).toEqual(EXPECTED_EXPIRED_TOKEN_ERROR)
         done()
       } catch (err) {
         done(err)
@@ -156,7 +156,7 @@ describe('API uploadSnapshot endpoint', () => {
     expect.hasAssertions()
     const onErrorCallback = (error) => {
       try {
-        expect(error).toEqual(EXPECTED_EXPIRED_TOKEN_ERROR)
+        expect(error).toEqual(ORIGINAL_EXPECTED_EXPIRED_TOKEN_ERROR)
         done()
       } catch (err) {
         done(err)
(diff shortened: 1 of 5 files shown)


ethan605 (changes requested):
Humm tbh the changes look just okay, but I don't like it, so many changes involved having a _temporary unblock_ of CI.

IMO we can do a bit differently:
- Instead of export `TEMP_EMPTY_ERROR`, let's export `EXPECTED_EXPIRED_TOKEN_ERROR` again
- Within `mockExpiredJwtAndResponse.js`, prepend the real `EXPECTED_EXPIRED_TOKEN_ERROR` with a `_` or `ORIGINAL_`
- Change `EXPECTED_EXPIRED_TOKEN_ERROR` to the empty values

kopijunkie:
> Humm tbh the changes look just okay, but I don't like it, so many changes involved having a _temporary unblock_ of CI.
> 
> IMO we can do a bit differently:
> 
> * Instead of export `TEMP_EMPTY_ERROR`, let's export `EXPECTED_EXPIRED_TOKEN_ERROR` again
> * Within `mockExpiredJwtAndResponse.js`, prepend the real `EXPECTED_EXPIRED_TOKEN_ERROR` with a `_` or `ORIGINAL_`
> * Change `EXPECTED_EXPIRED_TOKEN_ERROR` to the empty values and keep the inline documentation for the future fixes

Good suggestion! I wasn't very comfortable with the amount of changes for this too but couldn't think of a better way.

src/components/utils/__integrations__/onfidoApi/selfie.integration.js
```diff
@@ -155,7 +156,7 @@ describe('API uploadSnapshot endpoint', () => {
     expect.hasAssertions()
     const onErrorCallback = (error) => {
       try {
-        expect(error).toEqual(EXPECTED_EXPIRED_TOKEN_ERROR)
+        expect(error).toEqual(ORIGINAL_EXPECTED_EXPIRED_TOKEN_ERROR)
```
ethan605, line 159: Should this be `EXPECTED_EXPIRED_TOKEN_ERROR`?

kopijunkie: Oops. You're right!

kopijunkie: Not for `/snapshot` endpoint. That is still returning the original `expired_token` error response.

ethan605: Damn this is annoying 😅

ethan605 (approved):
LGTM!

stefaniacardenas (approved):
Is the Onfido API response breaking the SDK behaviour? This sounds like something that might be breaking several SDK versions, not only for the integrator but also the user.

src/components/utils/__integrations__/onfidoApi/selfie.integration.js
kopijunkie: Agreed. 😅

kopijunkie:
> Is the Onfido API response breaking the SDK behaviour? This sounds like something that might be breaking several SDK versions, not only for the integrator but also the user.

User will still see the generic error message on attempting upload with an expired JWT token (tried on Surge latest), so it's not breaking the SDK behaviour for end user I think. But you are right about it affected other SDK versions. Which is why I can't confirm the SDK behaviour hasn't been broken for sure.

PR #1651, opened 2021-12-10
stop taking snapshots when selfie is being taken

jpserra:
# Problem

In some rare cases, a concurrency problem seems to be happening when `useMultipleSelfieCapture` feature is enabled and we take periodic snapshots to compare with the Selfie where the snapshot and the Selfie are the same.

# Solution

Immediately stop taking snapshots once the Selfie is being taken to avoid that a snapshot is taken at the same time as the Selfie, causing the binaries to be exactly the same.

## Checklist


stefaniacardenas (commented):
LGTM. Can you add a changelog entry and a PR description?

kopijunkie (commented):
Change itself looks OK & I'm not very familiar with that specific feature as @stefaniacardenas  but I think the `useMultipleSelfieCapture` feature, if enabled, is reliant on snapshots being sent also. 🤔

jpserra:
@kopijunkie it does but this change doesn't prevent snapshots to be taken right? at most it will cause _less_ snapshots to be taken as it stops the interval sooner? but there should always be at least a snapshot (from the setTimeout)

kopijunkie:
@jpserra Yeah, you're right. There will be at least the initial 1 snapshot. Got the wrong idea from the wording of the PR title. Would suggest making this clear in the `CHANGELOG` entry, which should be created for this change also.

jpserra:
@kopijunkie @stefaniacardenas update PR description and added changelog entry

kopijunkie (approved):

stefaniacardenas (approved):

PR #1664, opened 2021-12-21
added missing await for async function calls

it-ony:
# Problem

In the tests, we call async functions without await. Webdriver / selenium is communicating with network calls to the browsers. Every call is async. If the result is not awaited than actually assertions might make, which are not true. 

# Solution

Fix calls to async functions with await statement. 

## Checklist


ethan605:
@it-ony, looks good to me in general, but I'm curious why the tests were running as expected (apart from the flaky situations) 🤔

it-ony:
I would call it luck :) Might also be, that the commands are processed in the order they're received. https://w3c.github.io/webdriver/webdriver-spec.html#protocol, so it might be that the server has a locking mechanism installed, that finishes the 1st request before the 2nd is processed and by this, we ensure indirectly the order.

PR #1679, opened 2022-01-25
Tony/prepare for new tests

it-ony:
# Problem

Tests should not check the content, but relay on unique id's per view.

# Solution

Introduce data-page-ids.

## Checklist


src/components/Router/CrossDeviceMobileRouter.tsx
```diff
@@ -99,7 +99,9 @@ export default class CrossDeviceMobileRouter extends Component<
     // the cross device flow without running nginx
     const url = props.urls.sync_url
 
-    const roomId = window.location.pathname.substring(3) || props.options.roomId
+    // TODO: review this change
+
+    const roomId = props.options.roomId || window.location.pathname.substring(3)
```
it-ony, line 104: When making the crossdevice working locally, it was always taking the input of the url in preference to the roomId passed as options. The or seems to be flipped. 

Please review @ethan605

src/components/crossDevice/CrossDeviceLink/index.js
```diff
@@ -246,8 +246,11 @@ class CrossDeviceLinkUI extends Component {
     const mobileUrl = cross_device_url || hosted_sdk_url
     // This lets us test the cross device flow locally and on Surge.
     // We use the same location to test the same bundle as the desktop flow.
+
+    // TODO: review this change
+
     return process.env.MOBILE_URL === '/'
-      ? `${window.location.origin}?link_id=${this.linkId}`
+      ? `${window.location.origin}${window.location.pathname}?link_id=${this.linkId}`
```
it-ony, line 253: The mobile url was always generated without the pathname. Running locally under http://localhost/somedir/ ended up in a link generated http://localhost/?link_id skipping the path.

/cc @ethan605

stefaniacardenas: In which cases did you need to use the pathname locally? Anyways, the change looks good to me

stefaniacardenas (approved):
Looks good to me

src/components/crossDevice/CrossDeviceLink/index.js
it-ony: For local testing without the need to rebuild the SDK. Before, we always assumed, that the cross origin url was served under http://host:port/.

src/components/Router/CrossDeviceMobileRouter.tsx
stefaniacardenas: As discussed, it should be safe to change this as we only need to fetch the room id from the URL when the roomId is not set in the store. So if roomId is null or undefined we can get it from the URL

PR #1680, opened 2022-01-26
Feature/sdk configuration service

nulrich:
# Problem

We want to use settings exposed on the SDK Configuration Service to control features of the SDK.

# Solution

Add SdkConfigurationServiceProvider at root of the application. It will call `/v3/sdk/configurations` API to retrieve the settings. Settings will be available through useSdkConfigurationService.

## Checklist


src/contexts/useSdkConfigurationService.tsx
```diff
@@ -0,0 +1,53 @@
...
+    </SdkConfigurationServiceContext.Provider>
+  ) : null
+}
+
+const useSdkConfigurationService = () => {
+  const configuration = useContext(SdkConfigurationServiceContext)
+
+  if (!configuration) {
+    throw new Error(`SDK Configuration wasn't initialized!`)
+  }
```
ethan605, line 48: Ditto here (`configuration` will have a default value)

ethan605 (changes requested):

src/contexts/useSdkConfigurationService.tsx
```diff
@@ -0,0 +1,53 @@
...
+type SdkConfigurationServiceProviderProps = {
+  children: ComponentChildren
+  url?: string
+  token?: string
+  sdkVersion?: string
+}
+
+const SdkConfigurationServiceContext = createContext<
+  SdkConfiguration | undefined
+>(undefined)
```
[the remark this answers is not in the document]
nulrich: In that case, I need introduce a 'loading' state to prevent the UI to be displayed until we get the settings from the API.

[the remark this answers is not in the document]
ethan605: Yeah sounds reasonable. With poor network conditions, this request can take a significant amount of time

stefaniacardenas (changes requested):
Some comments/questions:
* Loading screen should be displayed during request
* If the feature is not ready, for example, in case we need to wait for the v4 endpoint to be available, I would suggest putting this feature behind a webpack configuration flag, so that it only gets used if the configuration is enabled. Otherwise, hold off merging to development until ready
* Has the cross-device flow been handled?
* Is the feature always enabled? Can it be disabled on development?

src/contexts/useSdkConfigurationService.tsx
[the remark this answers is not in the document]
nulrich: I'm putting the Spinner (with autoFocusOnInitialScreenTitle support) used in the other screens instead of a blank page..

src/components/utils/onfidoApi.ts
```diff
@@ -382,3 +383,29 @@ export const sendAnalytics = (
...
+export const getSdkConfiguration = (
+  url: string,
+  token: string
+): Promise<SdkConfiguration> =>
+  new Promise((resolve, reject) => {
+    if (!process.env.SDK_SOURCE) {
+      throw new Error('SDK_SOURCE not defined')
+    }
+
+    const sdk_source = process.env.SDK_SOURCE.replaceAll('_', '-')
```
ethan605, line 396: We don't need to normalise from `onfido_web_sdk` to `onfido-web-sdk`. Our data pipeline has been living with the former value for ages 😅

src/contexts/useSdkConfigurationService.tsx
```diff
@@ -0,0 +1,63 @@
...
+  }, [url, token])
+
+  return configuration ? (
+    <SdkConfigurationServiceContext.Provider value={configuration}>
+      {children}
+    </SdkConfigurationServiceContext.Provider>
+  ) : (
+    <Fragment>{fallback}</Fragment>
+  )
+}
```
ethan605, line 57: Just the matter of personal taste:
```suggestion
  if (!configuration) {
    return <Fragment>{fallback}</Fragment>
  }
  
  return (
    <SdkConfigurationServiceContext.Provider value={configuration}>
      {children}
    </SdkConfigurationServiceContext.Provider>
  )
}
```

webpack.config.babel.js
```diff
@@ -176,6 +176,7 @@ const DEVELOPMENT_CONFIG = {
   PUBLIC_PATH: '/',
   MOBILE_URL: '/',
   RESTRICTED_XDEVICE_FEATURE_ENABLED: false,
+  SDK_CONFIGURATION: {},
```
ethan605, line 179: Please backfill this with default configuration of disabled multi-frame capture 🙏

webpack.config.babel.js
```diff
@@ -276,8 +277,7 @@ const baseConfig = {
     __dirname: false,
     setImmediate: false,
   },
-
-  devtool: PRODUCTION_BUILD ? 'source-map' : 'eval-cheap-source-map',
+  devtool: PRODUCTION_BUILD ? 'source-map' : 'cheap-module-source-map',
```
ethan605, line 280: Might be a change by accident 😄

src/contexts/useSdkConfigurationService.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  useEffect(() => {
+    if (!url || !token) {
+      return
+    }
+    getSdkConfiguration(url, token)
+      .then((configuration) => {
+        let mergedConfiguration = deepmerge(defaultConfiguration, configuration)
+
+        if (process.env.NODE_ENV === 'development') {
```
ethan605, line 38: Humm why?

webpack.config.babel.js
nulrich: 😅

webpack.config.babel.js
nulrich: It's planned in the MFC PR, but I can do it now.

ethan605: Ah okay then. You can keep this for a minimal diffs 🙏

src/components/utils/onfidoApi.ts
nulrich: So we need to update the Configuration Service which expects "onfido-web-sdk". Or I put back the hardcoded value.

ethan605: You should keep the `SDK_SOURCE` env var and update sdk-configuration-service to accept `onfido_web_sdk`

src/contexts/useSdkConfigurationService.tsx
ethan605: Oh no, I meant why only for `development`?

nulrich: * The default configuration will be defined in this file. It's the fallback configuration.
* In the webpackconf, it's only a way for devs to override a configuration value without touching the default value.

src/components/utils/onfidoApi.ts
nulrich: 👍

src/contexts/useSdkConfigurationService.tsx
ethan605: > In the webpackconf, it's only a way for devs to override a configuration value without touching the default value.

Well I guess we can simplify the _default_ `SDK_CONFIGURATION` at the front-end. This default value will be the fallback in case we are unable to load the remote configuration. So in each env, we'll expect a different value of this. But for the default value, we'll make it ... default, like no MFC, etc.

So IMO we'll be fine to have `SDK_CONFIGURATION` for `PROD_CONFIG` and propagate it to other envs

nulrich: 👍

src/contexts/useSdkConfigurationService.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+type SdkConfigurationServiceProviderProps = {
+  children: ComponentChildren
+  url?: string
+  token?: string
+  fallback?: ComponentChildren
+}
+
+const defaultConfiguration: SdkConfiguration = {}
+const envConfiguration = process.env.SDK_CONFIGURATION as SdkConfiguration
```
stefaniacardenas, line 15: Why is this an env variable and it's an empty object? Is this to allow us to pass the configurations we want at development time?

stefaniacardenas: These could only be changed at build time, though. Do we want that?

nulrich: * Default configuration is the fallback when we can't reach the server.
* envConfiguration, configuration in webpack, is the specific configuration per env, to override the server configuration (Ex: We may want to disable something specific in E2E).

stefaniacardenas: I see. Wouldn't it make more sense to expose a function to update the configuration at runtime for a specific env so you don't need to rebuild?

src/components/utils/sdkBackend.ts
```diff
@@ -44,11 +44,12 @@ export const postToBackend = (
 
   const endpoint = `${url}/validate_document`
 
-  const options = {
+  const options: HttpRequestParams = {
     payload,
     endpoint,
     token,
     contentType: 'application/json',
+    method: 'POST',
```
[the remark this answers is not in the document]
nulrich: 👍

src/index.tsx
```diff
@@ -117,9 +119,10 @@ const getContainerElementById = (containerId: string) => {
   )
 }
 
-export const init = (opts: SdkOptions): SdkHandle => {
+export const init = (opts: SdkOptions, conf?: SdkConfiguration): SdkHandle => {
```
stefaniacardenas, line 122: This would allow anyone to override the configurations. I think we should only allow this if the environment is `test` (so for surge demo app and for E2E tests)

stefaniacardenas (changes requested):

src/index.tsx
```diff
@@ -128,14 +146,15 @@ export const init = (opts: SdkOptions): SdkHandle => {
...
   } else if (options.containerId) {
     containerEl = getContainerElementById(options.containerId)
-    onfidoRender(options, containerEl)
+    onfidoRender(options, internalConfiguration, containerEl)
   }
 
   return {
     options,
+    internalConfiguration,
     setOptions(changedOptions) {
```
stefaniacardenas, line 158: I think we should update the setOptions function too

stefaniacardenas (changes requested):

stefaniacardenas (commented):
The PR looks good to me. Are we going to change the endpoint to point at v4 before merging?

nulrich:
The endpoint v4 is not ready, and I would like this code to finish MFC.
But I'll update it to v4 as soon as it's available.

nulrich:
Finally, the v4 is not ready. They advised me to stick on the v3.

stefaniacardenas (approved):

stefaniacardenas (approved):

PR #1686, opened 2022-02-03
Feature/error screens analytics

DannyvanderJagt:
# Problem

# Solution

## Checklist


src/Tracker/trackerData.tsx
```diff
@@ -503,4 +464,432 @@ export const analyticsEventsMapping = new Map<
...
     { eventName: 'WELCOME', properties: { event_type: 'screen' } },
   ],
+  [
+    'screen_document_front_confirmation_cutoff_detected',
+    {
+      eventName: 'DOCUMENT_CONFIRMATION_ERROR',
+      properties: {
+        event_type: 'view',
+        document_side: 'front',
+        ui_alerts: { cutoff: 'warning' },
```
stefaniacardenas, line 474: What exactly do we mean by 'warning'? This alert is not necessarily going to be a warning from UI POV. The user can only skip these alerts after two attempts so from their prospective this is an error. Just want to understand why we are sending "warning" or "error" for the UI alerts.

src/Tracker/trackerData.tsx
```diff
@@ -503,4 +464,432 @@ export const analyticsEventsMapping = new Map<
...
+      },
+    },
+  ],
+  [
+    'screen_document_back_confirmation_document_detection',
+    {
+      eventName: 'DOCUMENT_CONFIRMATION_ERROR',
+      properties: {
+        event_type: 'view',
+        document_side: 'front',
```
stefaniacardenas, line 612: This needs to be changed to 'back'
```suggestion
        document_side: 'back',
```

src/Tracker/trackerData.tsx
```diff
@@ -503,4 +464,432 @@ export const analyticsEventsMapping = new Map<
...
+      },
+    },
+  ],
+  [
+    'screen_document_back_confirmation_invalid_type',
+    {
+      eventName: 'DOCUMENT_CONFIRMATION_ERROR',
+      properties: {
+        event_type: 'view',
+        document_side: 'front',
```
stefaniacardenas, line 634: This needs to be changed to 'back'
```suggestion
        document_side: 'back',
```

stefaniacardenas (changes requested):

src/Tracker/trackerData.tsx
DannyvanderJagt: I think the initial reasoning was because it doesn't get the user stuck, as it only tries it twice, but I agree that it's received as an error by the user.

Thinking about it, I would think about a warning as a non interruptive message that they can ignore or dismiss and an error as something (temporary) blocking or interrupting their flow. What do you think?

There are a couple of "errors" defined as warnings or are dismissible, like camera inactivity (can be dismissible) and recording time out (with the action reset). I would label them as warnings for example.

stefaniacardenas: Yes, I agree. I would consider an error anything that is preventing the user from proceeding and cannot be dismissed

PR #1709, opened 2022-02-17
Tony/fix integration tests

it-ony:
# Problem

Integration tests where failing, as the assertion checks that the error is an expired token. The token we're using is from type `alg HS256`. Yesterdays api release will not let old tokens be used and only ones with type `HS512`. 

# Solution

Actually the assertion goes way to far as the tests should only test that with an outdated jw-token the error callback is triggered. I replaced the assertion. We can bring it back, if we have a new `HS512` token that is expired.

## Checklist


DannyvanderJagt (approved):
LGTM.  Maybe we could clean up the `console.log API error XMLHttpRequest {}` in the output. See `src/components/utils/onfidoApi.ts:357`

nulrich (approved):

PR #1712, opened 2022-02-21
Add compatibility with Salesforce

nulrich:
Ensure the code can run with Strict CSP and Locker.

# Problem

Salesforce Locker is redefining the window object: There's no prototype for `window` and all properties are owned.

For example, we are creating a proxy of `window` for Woopra. We retrieve all properties of `window` using:

```
for (const key in window) {
  propertyKeys.add(key)
}
```

But in Salesforce, some properties are now non-enumerable and owned. So they are missing in our proxy.

Also the version of PDFObject that we're using need some properties not supported by Salesforce.

# Solution

- We need to use `getOwnPropertyNames` to retrieve those properties.
- We need to inject the window proxy without using a global variable.
- We need to update PDFObject.

## Checklist


stefaniacardenas (approved):
Thanks for looking into this issue, it has been open for a while 🎉

zoeradkani (approved):

stefaniacardenas (approved):

PR #1746, opened 2022-03-02
WIP: SA-76: Ensure PoA and Document steps can be used together

dlewton:
# Problem

When a user chooses to use the Document step along with the PoA step, the data from one overwrites the other in the current development. Specifically, only the data from one of these steps is sent through to the onComplete callback.

# Solution

Rework the code so that the PoA step generates a separate key in the onComplete callback. This should ensure that the data from either step can be accessed within the onComplete callback.

## Checklist

- [yes] Has the CHANGELOG been updated?
- [yes] Has the README been updated?
- [?] Have any new automated tests been implemented or the existing ones changed?
- [?] Have any new manual tests been written down or the existing ones changed?


package.json
```diff
@@ -194,6 +194,7 @@
     "postcss-custom-media": "^7.0.8",
     "postcss-loader": "^3.0.0",
     "postcss-url": "^8.0.0",
+    "pre-commit": "^1.2.2",
```
it-ony, line 197: why is pre-commit added again? We just got rid of it.

src/components/utils/redux.ts
```diff
@@ -9,7 +9,9 @@ export const buildCaptureStateKey = (payload: {
 }): Exclude<keyof CaptureState, 'takesHistory'> => {
   const { method, side, variant } = payload
 
+  if (method === 'poa') return 'poa'
```
it-ony, line 12: let me suggest a map here

dlewton: I'm not sure I understand how to implement the suggestion of a map here. Currently, only the values of `poa` and `face` directory correspondent to a `method`. But, other methods depend on the values of `side` and `variant`.

dlewton:
An attempt to correct issues with pre-commit affecting my PR codebase through rebasing turned this PR into a mess. The best option forward appears to be to re-create a cleaner PR.

PR #1763, opened 2022-03-17
Remove unnecessary config from analytics

DannyvanderJagt:
# Problem
We're sending unnecessary config data with our analytics 

# Solution
Remove the unnecessary config and strip the locales from the config 

## Checklist

### DIFF
--- a/src/Tracker/onfidoTracker.tsx
+++ b/src/Tracker/onfidoTracker.tsx
@@ -8,6 +8,7 @@ import type { AnalyticsPayload, LegacyTrackedEventNames } from '~types/tracker'
 import { reduxStore } from 'components/ReduxAppWrapper'
 import { analyticsEventsMapping } from './trackerData'
 import { trackException } from './'
+import { pick } from '~utils/object'
 
 let currentStepType: ExtendedStepTypes | undefined
 let analyticsSessionUuid: string | undefined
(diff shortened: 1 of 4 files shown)


src/Tracker/onfidoTracker.tsx
```diff
@@ -51,6 +52,32 @@ const source_metadata = {
...
 const stepsArrToString = () => steps?.map((step) => step['type']).join()
 
+const removeKeys = [
+  'title',
+  'descriptions',
+  'nextButton',
+  'message',
+  'submessage',
+]
+const cleanStepsForConfig = (steps: StepConfig[]): StepConfig[] =>
```
stefaniacardenas, line 62: I think it would be better to pick only the configuration we want to include rather than removing those we don't want. Otherwise, any new step configuration will automatically be tracked and we would easily lose track of what we're sending.

src/Tracker/onfidoTracker.tsx
```diff
@@ -53,6 +54,40 @@ const source_metadata = {
...
+  steps.map(
+    (step): StepConfig => {
+      const options = pick(
+        { ...step.options } as Partial<Record<string, unknown>>,
+        keys
+      )
+      const newStep: StepConfig = { ...step }
+
+      if (Object.keys(options).length) {
+        newStep.options = options
```
stefaniacardenas, line 82: Ideally we should convert the keys to snake case before sending the Looker, so they are more readable

DannyvanderJagt: Good to know. I converted the first level of `step.options` to snake_case. The second level properties are already in snake_case according to our documentation.

stefaniacardenas (approved):

PR #1830, opened 2022-04-28
feat: reposition "submit" button for profile data screen

josokinas:
# Problem

Reposition "submit" button for profile data screen so that it always at the bottom of the form, and seen only when screen scrolled down.

# Solution

Place button for form submission directly after form, rather than in "actions" section of a screen.

### DIFF
--- a/src/components/Capture/ProfileData.tsx
+++ b/src/components/Capture/ProfileData.tsx
@@ -76,17 +76,8 @@ const ProfileData = ({
     nextStep()
   }
 
-  const actions = (
-    <Button
-      onClick={handleSubmit}
-      className={classNames(theme['button-centered'], theme['button-lg'])}
-    >
-      {translate('profile_data.button_submit')}
-    </Button>
-  )
-
   return (
-    <ScreenLayout actions={actions}>
+    <ScreenLayout>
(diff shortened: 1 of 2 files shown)


src/components/Capture/ProfileData.tsx
```diff
@@ -122,6 +113,16 @@ const ProfileData = ({
...
+      <Button
+        onClick={handleSubmit}
+        className={classNames(
+          theme['button-centered'],
+          theme['button-lg'],
+          style['submit-button']
+        )}
+      >
+        {translate('profile_data.button_submit')}
+      </Button>
```
zoeradkani, line 125: Just wondering if it is sitting across the bottom of the page rightly if we take it out of screen layout

zoeradkani (approved):

src/components/Capture/ProfileData.tsx
josokinas: It is no longer, no.

E.g.:
Screenshot 2022-04-28 at 11 44 20

Screenshot 2022-04-28 at 11 43 55

Screenshot 2022-04-28 at 11 44 30

zoeradkani: as long as it is the same in the design then.

PR #1852, opened 2022-05-05
fix: add "format" as helper text for date of birth field

josokinas:
# Problem

Date of birth field needs to have a format described as helper text.

# Solution

When building helper text for a field, use a custom format one for "dob" case.

Before:

Screenshot 2022-05-05 at 14 20 04

After:

Screenshot 2022-05-05 at 14 20 38

### DIFF
--- a/src/components/Capture/ProfileData.tsx
+++ b/src/components/Capture/ProfileData.tsx
@@ -265,9 +265,20 @@ const getTranslatedFieldHelperText = (
 ) => {
   const specificTranslation = translateSpecific('helper_text', type, country)
 
-  return specificTranslation ? (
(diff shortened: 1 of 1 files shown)


src/components/Capture/ProfileData.tsx
```diff
@@ -265,9 +265,20 @@ const getTranslatedFieldHelperText = (
...
+    return (
+      <HelperText>
+        {translate(`profile_data.${specificTranslation}`)}
+      </HelperText>
+    )
+  }
+
+  switch (type) {
+    case 'dob':
+      return <HelperText>MM / DD / YYYY</HelperText>
```
zoeradkani, line 278: Why do we change the labels for non US but keep the dates as American format for all languages?

zoeradkani (approved):

PR #1861, opened 2022-05-10
📈 Update Biometrics analytics events

warrenseine:
# Problem

See BIO-4326 ticket.

# Solution

New events have been added where possible according to the spreadsheet linked in the ticket.

However, it does not match 100%:
- "back" and "close" buttons cannot be tracked for each component, so generic events `NAVIGATION_BACK_BUTTON_CLICKED ` and `NAVIGATION_CLOSE_BUTTON_CLICKED` with the `step` property are emitted instead
- tracked components in nested screens impose a screen name hierarchy so `FACE_SELFIE_CAMERA_ACCESS_*` became `FACE_SELFIE_CAPTURE_CAMERA_ACCESS_*`
- `FACE_SELFIE_REVIEW` already exists as `FACE_SELFIE_CONFIRMATION`, so all `FACE_SELFIE_REVIEW_*` are instead named `FACE_SELFIE_CONFIRMATION_*` (c.f. previous point, and for consistency)

## Checklist


src/components/CameraPermissions/withPermissionsFlow.tsx
```diff
@@ -92,7 +93,10 @@ export default <P extends Props>(
       return (
         <PermissionsPrimer
           {...{ trackScreen, audio }}
-          onNext={this.setPermissionsPrimerSeen}
+          onNext={trackClick(
```
DannyvanderJagt, line 96: This will go for all the screens that use it, not only face selfie. Can you replace it with a generic name?

src/components/CameraPermissions/Recover/index.tsx
```diff
@@ -38,7 +38,10 @@ const Recover: FunctionComponent<Props> = ({ translate, audio }) => {
     <Button
       variant="primary"
       className={classNames(theme['button-centered'], theme['button-lg'])}
-      onClick={() => window.location.reload()}
+      onClick={trackClick(
+        () => window.location.reload(),
+        'screen_face_selfie_capture_camera_access_denied_refresh_button'
```
DannyvanderJagt, line 43: This will go for all screens that use it, not only face selfie. Can you replace it with a generic name?

src/components/CameraPermissions/Primer/index.tsx
```diff
@@ -70,4 +70,4 @@ const Permissions: FunctionComponent<Props> = ({
   )
 }
 
-export default trackComponent(localised(Permissions))
+export default trackComponent(localised(Permissions), 'camera_access')
```
DannyvanderJagt, line 73: There are more flows that use this, you will need to add all those event to the code.

Missing:
`screen_document_document_video_capture_camera_acces`
`screen_face_face_video_capture_camera_access`

src/components/CameraPermissions/Recover/index.tsx
```diff
@@ -71,4 +74,4 @@ const Recover: FunctionComponent<Props> = ({ translate, audio }) => {
   )
 }
 
-export default trackComponent(localised(Recover))
+export default trackComponent(localised(Recover), 'camera_access_denied')
```
DannyvanderJagt, line 77: Same here, used by multiple screens. Need to add the events.

Missing:
`screen_face_face_video_capture_camera_access_denied`
`screen_document_document_video_capture_camera_access_denied`

src/components/Capture/index.ts
```diff
@@ -26,17 +26,21 @@ export const PoACapture = appendToTracking(
 )
 
 export const SelfieCapture = appendToTracking(
-  withCaptureVariant(Face, {
-    requestedVariant: 'standard',
-    pageId: 'SelfieCapture',
-  }),
+  trackComponent(
```
DannyvanderJagt, line 29: This is not right, please remove it. Which event did you intent to fire?

The events inside `<Face/>` are fired by specific components inside. It's a bit confusing.

src/components/Capture/index.ts
```diff
@@ -26,17 +26,21 @@ export const PoACapture = appendToTracking(
...
+    })
+  ),
   'selfie_capture'
 )
 export const FaceVideoCapture = appendToTracking(
-  withCaptureVariant(Face, {
-    requestedVariant: 'video',
-    pageId: 'FaceVideoCapture',
-  }),
+  trackComponent(
```
DannyvanderJagt, line 38: Same here, `trackComponent`

src/components/Confirm/Confirm.tsx
```diff
@@ -497,6 +497,8 @@ class Confirm extends Component<ConfirmProps, ConfirmState> {
   onRetake = () => {
     const { actions, previousStep } = this.props
 
+    sendEvent('screen_face_selfie_confirmation_retake_button')
```
DannyvanderJagt, line 500: Also this screen is used by almost all flows. We can't add such specific event name like this as it will fire for most flows.

src/components/Confirm/Confirm.tsx
```diff
@@ -513,6 +515,8 @@ class Confirm extends Component<ConfirmProps, ConfirmState> {
   }
 
   onConfirm = () => {
+    sendEvent('screen_face_selfie_confirmation_upload_button')
```
DannyvanderJagt, line 518: Same

DannyvanderJagt (changes requested):
Needs changes. Let me know if you need any help on this.

warrenseine:
Noticed a bug on the "refresh" button (some code didn't make it through the commit rebases). Will fix and update PR.

src/components/CameraError/index.tsx
```diff
@@ -31,7 +32,10 @@ export default class CameraError extends Component<Props, State> {
 
   trackFallbackClick = (): void => {
     if (this.props.error.type === 'warning') {
-      this.props.trackScreen('fallback_triggered')
+      const properties: AnalyticsEventProperties = {
```
DannyvanderJagt, line 35: 1) Don't send an empty error object if there is not error
2) These kinds of "errors" are send as "ui_alerts: { [name]: [type] }" See Google sheets (for example `FACE_VIDEO_CAPTURE_ERROR`)

DannyvanderJagt (changes requested):
Overall I like your improvements 👍 Only the fallback errors need to be `ui_alerts` I think.

Events that are fired but missing from `trackData.tsx`:
- `screen_document_front_confirmation_upload_button_clicked`
- `screen_document_back_confirmation_upload_button_clicked`
- `screen_face_face_video_confirmation_upload_button_clicked`

src/components/CameraError/index.tsx
warrenseine: 1. I'm not sure I understand. In this branch, error cannot be empty 🤔
2. Yes, I did notice that and it's already filterable in our SDK Analytics Looker Explore, which is nice. However, at this point, the only prop I have is an `ErrorProp`, whereas `ui_alerts` expects a `UIAlerts` map. Would it be fine if I implement a function to convert a `ErrorProp` to a `UIAlerts` map? I could then get rid of this `error` property I'm adding and match the existing convention that you suggest. The other option would be to rename all the `ErrorNames` values, but I'm afraid this had non-trivial impact.

DannyvanderJagt (approved):
Looks good! 👍

PR #1909, opened 2022-06-10
Hide User Consent when consent already given

nulrich:
# Problem

When User Consent is the first step, it wasn't hide when the consent was
already given. 

This, because we were not waiting for User Consent step to be injected, so the history
router was pointing to the wrong step index.

# Solution

Now, we display the spinner until the user consent is injected, and then
we display history router with the proper list of steps. History Router
will also ensure to skip Steps flagged with `skip: true`.

Also, cleaning types to avoid the dependency between `StepsRouterProps`
and the Step Hook. And, adding a `loading` boolean instead of the
`status` variable.

## Checklist

- [X] Have any new automated tests been implemented or the existing ones changed?

### DIFF
--- a/src/components/Router/HistoryRouter.tsx
+++ b/src/components/Router/HistoryRouter.tsx
@@ -1,6 +1,6 @@
-import { h } from 'preact'
+import { Fragment, h } from 'preact'
 
-import { buildStepFinder, findFirstIndex } from '~utils/steps'
+import { buildStepFinder, findFirstEnabled, findFirstIndex } from '~utils/steps'
 import { buildComponentsList } from './StepComponentMap'
 import StepsRouter from './StepsRouter'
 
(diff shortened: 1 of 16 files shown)


src/components/Router/HistoryRouter.tsx
```diff
@@ -36,6 +37,30 @@ const formattedError = ({ response, status }: ParsedError): FormattedError => {
...
+        <p>{error}</p>
+        <p>Please try reloading the app, and try again.</p>
+      </div>
+    )
+  }
+
+  return <HistoryRouter {...props} {...useStepsProps} steps={steps} />
+}
+
 export const HistoryRouter = (props: HistoryRouterProps) => {
```
osolliec, line 64:
```suggestion 
const HistoryRouter = (props: HistoryRouterProps) => {
```
It looks like it's not used outside this file now ?

osolliec (changes requested):
one nitpick comment

src/components/Router/HistoryRouter.tsx
DannyvanderJagt: Going to merge this as it's holding up the release. This can be fixed if needed in a new PR

nulrich: I will add this in an other PR

zoeradkani (approved):

PR #1915, opened 2022-06-13
Integrate new facial similarity product

warrenseine:
# Problem

Integrate our new facial similarity product into the Onfido Web SDK.

# Solution

The solution is described in [our internal wiki](https://wiki.onfido.net/display/ENG/Web+Solution).

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -233,6 +233,7 @@
     "yn": "^3.0.0"
   },
   "dependencies": {
+    "@onfido/active-video-capture": "^0.0.1",
     "@sentry/browser": "^6.19.7",
     "blueimp-load-image": "~2.29.0",
     "classnames": "~2.2.5",
--- a/src/components/ActiveVideo/ActiveVideo.tsx
+++ b/src/components/ActiveVideo/ActiveVideo.tsx
@@ -1,0 +1,57 @@
+import { ActiveVideoCapture, LivenessError } from '@onfido/active-video-capture'
+import { h, FunctionComponent } from 'preact'
+import { useState } from 'preact/hooks'
+import { localised } from '~locales'
+import type { WithLocalisedProps } from '~types/hocs'
+import type { StepComponentBaseProps } from '~types/routers'
+
+type Props = StepComponentBaseProps & WithLocalisedProps
+
+const ActiveVideo: FunctionComponent<Props> = (props) => {
+  const { nextStep } = props
+  const [videoPayload, setVideoPayload] = useState<Blob | null>(null)
+  const [success, setSuccess] = useState<boolean>(false)
+  // FIXME: Move to src/locales/en_US/en_US.json
+  const localisedText = {
+    avc_face_alignment: {
+      feedback_move_back: 'Move back',
+      feedback_move_closer: 'Move closer',
+      feedback_no_face_detected: 'Face not detected',
+      feedback_not_centered: 'Face not centered',
+      title: 'Position your face in the frame',
+    },
+    avc_face_capture: {
+      alert: {
+        timeout_body: 'You have up to 15 seconds to complete the recording',
+        timeout_button_primary: 'Retry',
+        timeout_title: 'Sorry, we have to restart the recording',
+        too_fast_body: 'Please retry and turn your head slower',
+        too_fast_button_primary: 'Retry',
+        too_fast_title: 'You turned your head too fast',
+      },
+      title: 'Turn your head slowly to both sides',
+      title_completed: 'Recording complete',
+    },
+  }
+
+  // See: https://github.com/preactjs/preact/issues/2748
+  return (
+    <ActiveVideoCapture
+      debug={true}
+      options={{
+        language: localisedText,
+      }}
+      onError={(event: LivenessError) => console.error(event)}
+      onSuccess={({ videoPayload }) => {
+        if (!success) {
+          setVideoPayload(videoPayload)
+          setSuccess(true)
+        }
+
+        nextStep()
+      }}
+    />
+  )
+}
+
+export default localised(ActiveVideo)
--- a/src/components/Router/StepComponentMap.tsx
+++ b/src/components/Router/StepComponentMap.tsx
@@ -29,6 +29,7 @@ import CrossDeviceClientIntro from 'components/crossDevice/ClientIntro'
 import ClientSuccess from '../crossDevice/ClientSuccess'
 import CrossDeviceIntro from '../crossDevice/Intro'
 import FaceVideoIntro from '../FaceVideo/Intro'
+import LazyActiveVideo from '../ActiveVideo/Lazy'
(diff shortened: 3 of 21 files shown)


src/components/Motion/Motion.tsx
```diff
@@ -0,0 +1,50 @@
+import { ActiveVideoCapture } from '@onfido/active-video-capture'
```
warrenseine, line 1: Probably needs to be lazy loaded as it adds 2 MB to the final bundle.

webpack.config.babel.js
```diff
@@ -280,7 +280,7 @@ const baseConfig = {
     __dirname: false,
   },
 
-  devtool: PRODUCTION_BUILD ? 'source-map' : 'eval-cheap-module-source-map',
+  devtool: PRODUCTION_BUILD ? 'source-map' : 'cheap-module-source-map',
```
warrenseine, line 283: For some reason, the app crashes in development builds when using `eval` after the `active-video-capture` library is loaded. The issue is identical to what's described [here](https://stackoverflow.com/a/69200112). The only downside [appears to be](https://webpack.js.org/configuration/devtool/) that the first build is slower (didn't notice any difference in practice).

src/components/Motion/Motion.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { h, FunctionComponent } from 'preact'
+import { useRef, useState } from 'preact/hooks'
+import { localised } from '~locales'
+import type { WithLocalisedProps } from '~types/hocs'
+import type { StepComponentBaseProps } from '~types/routers'
+
+type Props = StepComponentBaseProps & WithLocalisedProps
+
+const VideoRecording = ({ videoFile }: { videoFile: Blob | undefined }) => {
```
warrenseine, line 10: It's a debugging helper, it should be removed.

src/components/Router/StepsRouter.tsx
```diff
@@ -65,6 +66,9 @@ class StepsRouter extends Component<StepsRouterProps> {
       ? hideOnfidoLogo
       : globalUserOptions.enterpriseFeatures?.hideOnfidoLogo && hideOnfidoLogo
 
+    // Motion Web SDK already includes a footer
+    const hideFooterLogic = step.type === 'motion'
```
nulrich, line 70: I feel that there are too many `step.type === 'motion'`. We could leverage `PrivateStepConfig` and having a more generic way to handle the specific display (transparent + edgeToEdge + hideFooter). WDYT?

warrenseine: I was not aware of this interface. Yeah, I think that'd work!

src/types/steps.ts
```diff
@@ -5,6 +5,7 @@ const STEP_POA = 'poa'
 const STEP_FACE = 'face'
 const STEP_COMPLETE = 'complete'
 const STEP_AUTH = 'auth'
+const STEP_MOTION = 'motion'
```
warrenseine, line 8: The step is called `ACTIVE_VIDEO_CAPTURE` in the Android SDK, so maybe I'll have to keep the code name here too for consistency.

package.json
```diff
@@ -233,6 +233,7 @@
     "yn": "^3.0.0"
   },
   "dependencies": {
+    "@onfido/active-video-capture": "file:../active-video-capture",
```
jpserra, line 236: where is this coming from?

src/components/Router/StepComponentMap.tsx
```diff
@@ -132,6 +133,7 @@ const buildCaptureStepComponents = (
   const faceStep = findStep('face')
   const documentStep = findStep('document')
   const dataStep = findStep('data')
+  const motionStep = findStep('motion')
```
jpserra, line 136: "motion" should be just a "variant" for the face step and not a separate step no?

adding to that, we don't want to support motion explicitly, but instead use or A/B approach to convert selfie/video to it so I guess it would be perhaps inside the `buildFaceComponents` that we would make that A/B logic and return either selfie/video or motion?

warrenseine: Yeah, this is an outdated MR. I have changes to push that reflect how it's done in the other SDKs: the step will be private and will replace dynamically the `face` step.

package.json
jpserra: makes sense, I figured that was what was happening but wanted to double check in case there was some magic thing I wasn't aware of 👌

warrenseine: Updated to use the published package.

src/components/ActiveVideo/ActiveVideo.tsx
```diff
@@ -0,0 +1,57 @@
...
+    avc_face_alignment: {
+      feedback_move_back: 'Move back',
+      feedback_move_closer: 'Move closer',
+      feedback_no_face_detected: 'Face not detected',
+      feedback_not_centered: 'Face not centered',
+      title: 'Position your face in the frame',
+    },
+    avc_face_capture: {
+      alert: {
+        timeout_body: 'You have up to 20 seconds to complete the recording',
```
jpserra, line 25: should we change this right away to 15? I know the component it self still allows 20 but we're gonna have to change that to match Android & iOS behaviour

src/contexts/useActiveVideo.tsx
```diff
@@ -0,0 +1,42 @@
+import { useCallback } from 'preact/compat'
+import useSdkConfigurationService from '~contexts/useSdkConfigurationService'
+import { StepConfig } from '~types/steps'
+
+const useActiveVideo = () => {
+  const { experimental_features } = useSdkConfigurationService()
+
+  const enabled = experimental_features?.enable_liveness_experiment ?? false
```
jpserra, line 8: `enable_liveness_experiment` is deprecated, lets use `experimental_features?.motion_experiment?.enabled` which is the new one

src/components/ActiveVideo/ActiveVideo.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useState } from 'preact/hooks'
+import { localised } from '~locales'
+import type { WithLocalisedProps } from '~types/hocs'
+import type { StepComponentBaseProps } from '~types/routers'
+
+type Props = StepComponentBaseProps & WithLocalisedProps
+
+const ActiveVideo: FunctionComponent<Props> = (props) => {
+  const { nextStep } = props
+  const [videoPayload, setVideoPayload] = useState<Blob | null>(null)
```
nulrich, line 12: `videoPayload` is not used. Do you really need this state ?

warrenseine: Yeah, that's the output of the step, it's essential 😅 We just don't upload it yet, but this MR is already big enough. It will be used in a follow-up MR.

nulrich (approved):

PR #1923, opened 2022-06-20
✨ Improve the demo app

warrenseine:
# Problem

In the demo app, I would like to add a button to turn on/off a feature that is normally provided by the SDK configuration service. Using the `overrideConfiguration` option, I can set the new value at initialisation time, but no change will be taking into account later.

# Solution

Remove the local state in `useSdkConfigurationService` and rely on props instead.

## Checklist

### DIFF
--- a/src/contexts/useSdkConfigurationService.tsx
+++ b/src/contexts/useSdkConfigurationService.tsx
@@ -36,13 +36,11 @@ export const SdkConfigurationServiceProvider = ({
   url,
   token,
   fallback,
-  overrideConfiguration = {},
+  overrideConfiguration,
 }: SdkConfigurationServiceProviderProps) => {
   const [configuration, setConfiguration] = useState<
     SdkConfiguration | undefined
   >(undefined)
-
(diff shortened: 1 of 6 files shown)


nulrich:
@warrenseine What is you use case? 

We were planning to remove the `overrideConfiguration`. It has been added some time ago to mock configuration in the Selenium tests, but now we can do it directly with the mock server.

warrenseine:
> @warrenseine What is your use case?

Activating an SDK configuration flag from the demo app, for manual testing and for bug bashing. Maybe we could just activate it for the test client if we want to get rid of the property.

warrenseine:
In the end, I'd prefer to merge this change anyway. During the latest bug bash, users thought they had the feature turned on after toggling a checkbox but they had to reload the page. When the new demo app is ready, ping me and I'll make the necessary changes to get rid of my use case of  `overrideConfiguration`.

I've added other commits in this PR, so feel free to review them all.

nulrich (approved):

PR #1931, opened 2022-06-24
feat: collect country of residence first, before other profile data

josokinas:
# Problem

When collecting profile data, we don't know "country" before asking to enter other info e.g. date of birth. So we can't format to a specific country one, e.g. US format is different from UK.

# Solution

Collect address first, then other data.


zoeradkani (approved):

src/components/Capture/ProfileData.tsx
```diff
@@ -92,6 +94,7 @@ const ProfileData = ({
             type={type as FieldComponentProps['type']}
             value={value as FieldComponentProps['value']}
             selectedCountry={formData.country}
+            disabled={(disabledFields || []).includes(type)}
```
zoeradkani, line 97: the dropdown is disabled but the value is not set

josokinas: Fixed 👍

PR #1957, opened 2022-07-12
fix:pass welcome config for workflow welcome screen

zoeradkani:
# Problem
Workflow welcome screen is going to be configureable shortly. Some clients request to be able to Edit the **title** and **button label** via the config of SDK. this is a temp solution to pass the configuration on.

# Solution

## Checklist

- [ n/a] Has the CHANGELOG been updated?
- [n/a ] Has the README been updated?
- [n/a ] Has the CONTRIBUTING doc been updated?
- [n/a ] Has the RELEASE_GUIDELINES been updated?
- [n/a ] Has the TESTING_GUIDELINES been updated?
- [n/a ] Has the MIGRATION doc been updated for any MAJOR breaking changes?
- [n/a ] Has the MIGRATION doc been updated for any MINOR breaking changes, including any translation strings or keys changes?
- [ n/a] Have any new automated tests been implemented or the existing ones changed?
- [n/a ] Have any new manual tests been written down or the existing ones changed?
- [n/a ] Have any new strings been translated or the existing ones changed?


src/index.tsx
```diff
@@ -72,12 +72,17 @@ const formatOptions = ({
...
+    steps?.map(formatStep).find((i) => i.type === 'welcome'),
+  ] || [{ type: 'welcome' }]
+
   return {
     ...otherOptions,
     smsNumberCountryCode: validateSmsCountryCode(smsNumberCountryCode),
     useWorkflow,
     steps: useWorkflow
-      ? defaultSteps.map(formatStep)
+      ? (welcomeStep || defaultSteps).map(formatStep)
```
DannyvanderJagt, line 85: If `defaultSteps` is suppose to be a fallback, then this doesn't work as welcome step always returns an array (`[undefined]`).

DannyvanderJagt (changes requested):

src/index.tsx
zoeradkani: @DannyvanderJagt very good point, I actually thought I had a condition to default the welcome step, update it in the block above, not so clean. what do you think?

DannyvanderJagt: Not sure what the logic here should be?

This will always use `welcomeStep` (as it's always populated by either the welcome step from `steps` or `[{ type: 'welcome' }]` and always ignore `defaultSteps`. For auth this will therefore remove the step `auth`.

If we don't use auth with workflow, then best to remove the `defaultSteps` from line 85 to clean it up. 

If you just want to update the welcome step if it's available in `Onfido.init` then something like this should work:

```tsx
 // @ts-ignore TODO: quick fix, remove this whole block when the welcome screen is configured via workflow
  const welcomeStep =
    steps?.map(formatStep).find((i) => i.type === 'welcome') || 'welcome'

  const defaultSteps: (StepTypes | StepConfig)[] =
    process.env.SDK_ENV === 'Auth'
      ? [welcomeStep, 'auth', ...mandatorySteps]
      : [welcomeStep, ...mandatorySteps]

  return {
    ...otherOptions,
    smsNumberCountryCode: validateSmsCountryCode(smsNumberCountryCode),
    useWorkflow,
    steps: useWorkflow
      ? defaultSteps.map(formatStep)
      : (steps || defaultSteps)
          .map(formatStep)
          .map(injectPrivateConfig)
          .filter(({ type }) => !internalSteps.includes(type)),
  }
```

zoeradkani: @DannyvanderJagt gotcha!!!! I totally discounted Auth ( in reality I think it is not part of workflow anyway) , in my head the polling would start after welcome screen ( at the moment welcome screen mandatory for workflow) so we always start with welcome screen. 
Thank you very much for your brilliant feedback.

DannyvanderJagt (approved):

PR #1985, opened 2022-07-27
Display an error when token is expired

nulrich:
# Problem

Currently, when the JWT token is expired:
- If user consent page is displayed, the applicant will get an error message only after accepting the conditions.
- If user consent is not displayed, the applicant will get an error message only after uploading a document

# Solution

The first call made by the SDK is /configuration. So now, in case of an error from the API, we only use the fallback configuration if the token is NOT expired. Otherwise we display a message to the applicant.

## Checklist

- [X] Have any new automated tests been implemented or the existing ones changed?


DannyvanderJagt:
FYI: I've made a simple network layer, which all our api requests will funnel through. We can use that to catch the first api request that returns the token invalid error, instead of relying on the sdk configuration service being the first.

https://github.com/onfido/onfido-sdk-ui/pull/1949 (network layer)
https://github.com/onfido/onfido-sdk-ui/pull/1951 (use it everywhere)

It's still in acceptance, but I can merge it to help this use case. Or if we're in a rush, we can merge this and change it later.

nulrich:
@DannyvanderJagt if you can merge your PR I'll use it 👍

DannyvanderJagt (approved):
Looks good 👍 It will work fine for now.

The network layer is merged, but in retrospect I have other prio's for this release so we can use it later.

PR #1987, opened 2022-07-28
Feature: Add new trackPropertiesBeforeMount property to WithTrackingProps

osolliec:
# Problem

Today there is no easy way to add dynamic properties to analytics events for class components.

# Solution

Add a new trackPropertiesBeforeMount method that can be defined on the tracked component, see commit "Use it in a class component" to see how it's used
```
trackPropertiesBeforeMount={() => ({ myProp: 'val' })}
```

## Checklist


src/components/Capture/Document.tsx
```diff
@@ -252,7 +252,10 @@ const Document = (props: Props) => {
       throw new Error('documentType not provided')
     }
 
-    if (sdkConfiguration.experimental_features?.enable_multi_frame_capture) {
+    if (
+      false &&
```
DannyvanderJagt, line 256: Should this be here?

src/Tracker/index.tsx
```diff
@@ -126,24 +126,14 @@ const trackComponent = <P extends WithTrackingProps>(
   class TrackedComponent extends Component<P> {
     callback?: () => AnalyticsDynamicProperties = undefined
     componentDidMount() {
-      const properties =
-        typeof this.callback === 'function' ? this.callback() : undefined
+      const properties = this.props.trackPropertiesBeforeMount
```
DannyvanderJagt, line 129: Would add `typeof this.props.trackPropertiesBeforeMount === 'function'` to be safe

src/components/Capture/Document.tsx
osolliec: removed

src/Tracker/index.tsx
osolliec: Do you think it can be changed at runtime ? Because at compile time, it will be enforced with the type WithTrackingProps:
```
export type TrackPropertiesBeforeMountCallback = () => Record<string, unknown>

export type WithTrackingProps = {
  trackScreen: TrackScreenCallback
  trackPropertiesBeforeMount?: TrackPropertiesBeforeMountCallback
}
```

DannyvanderJagt (approved):

PR #1988, opened 2022-07-28
feat(orc-1253): SSN input autocomplete with mask

acMaiaOnf:
# Problem

Currently people need to write the complete SSN with `-` for it to be valid
When copying it, everything gets copied

# Solution

automatically add the SSN mask when typing
remove the mask when copying from the input

### DIFF
--- a/src/components/Capture/ProfileData.tsx
+++ b/src/components/Capture/ProfileData.tsx
@@ -17,6 +17,7 @@ import theme from '../Theme/style.scss'
 import { CountrySelector } from './CountrySelector'
 import { StateSelector } from './StateSelector'
 import { DateOfBirthInput, getMaxDay } from './DateOfBirthInput'
+import { SSNInput } from './SSNInput'
 import style from './style.scss'
 
 import type { StepComponentDataProps, CompleteStepValue } from '~types/routers'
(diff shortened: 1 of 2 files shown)


kai-moseley-onfido (approved):
One thing I'm always cautious about (so worth double checking) is that when we modify an input's value it can cause cursor shenanigans when typing in the middle of the existing word. If that's working as it should, then lgtm.

src/components/Capture/ProfileData.tsx
```diff
@@ -23,6 +23,7 @@ import type { StepComponentDataProps, CompleteStepValue } from '~types/routers'
 import type { StepOptionData } from '~types/steps'
 import type { WithLocalisedProps } from '~types/hocs'
 import type { TranslateCallback } from '~types/locales'
+import { SSNInput } from './SSNInput'
```
josokinas, line 26: [Minor] Better to put import next to code imports (not type).

src/components/Capture/ProfileData.tsx
```diff
@@ -291,12 +292,10 @@ const getFieldComponent = (
       return <Input {...props} type="text" style={{ width: space(22) }} />
     case 'ssn':
       return (
-        <Input
+        <SSNInput
           {...props}
           placeholder={'999-99-9999'}
-          pattern={'d{3}-d{2}-d{4}'}
           maxLength={11}
```
josokinas, line 298: IMO better to put `maxLength` inside custom component also? And probably `placeholder` too?

josokinas (approved):

PR #2009, opened 2022-08-03
♻️ Convert `CameraError` to a function component

warrenseine:
# Problem

There's a warning about using `setState` in `componentDidUpdate`.

# Solution

Instead of working around it with `getDerivedStateFromProps`, the component was migrated to a function component with hooks.

## Checklist

### DIFF
--- a/src/components/CameraError/index.tsx
+++ b/src/components/CameraError/index.tsx
@@ -1,4 +1,4 @@
-import { h, Component } from 'preact'
+import { h, FunctionComponent } from 'preact'
 import Error from '../Error'
 import classNames from 'classnames'
 import { parseTags } from '~utils'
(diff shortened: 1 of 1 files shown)


nulrich (approved):

src/components/CameraError/index.tsx
```diff
@@ -18,72 +19,60 @@ type Props = {
...
-      isDismissible,
-      trackScreen,
-    } = this.props
+  }, [error, trackScreen])
 
-    if (this.state.isDimissed) {
-      return null
-    }
+  useEffect(() => {
+    setIsDismissed(false)
```
osolliec, line 43: shouldn't we test the error name equality ? (Unsure if the case happens that a new error object pops up with the same name)

warrenseine: You're right. I don't know if it happens, but if the component is re-rendered and the reference to the `error` object changes, it will dismiss the error message.

I can just replace `error` with `error.name` in the dependency array to get the exact same behaviour as before.

osolliec (approved):

PR #2010, opened 2022-08-03
🚨 Remove additional @ts-ignore

warrenseine:
# Problem

Warnings.

# Solution

Fixing warnings.

## Checklist


src/types/tracker.ts
```diff
@@ -150,6 +150,9 @@ export type LegacyTrackedEventNames =
   | 'Error response from onSubmitSelfie'
   | 'Error response from onSubmitVideo'
   | 'Error response from onSubmitDocument'
+  | 'Success response from onSubmitSelfie'
```
DannyvanderJagt, line 153: Are those analytics events? Then we should add them to trackerData too and fire them.

DannyvanderJagt (approved):
Thanks for the cleanup 👍

osolliec (approved):

src/types/tracker.ts
warrenseine: They are fired already [here](https://github.com/onfido/onfido-sdk-ui/blob/fix/warnings/src/components/Confirm/Confirm.tsx#L400). What you're saying is that it's not actually sent because it's not mapped in `trackerData`?

I have no idea what these events are for. Maybe it's historical and nobody reads them, so nobody cared when we migrated to a more structured/standardised approach?

PR #2020, opened 2022-08-06
🏗️ Fix Motion navigation bar integration

warrenseine:
The hack introduced in the Web SDK to allow the navigation bar to appear transparent and the component to take the full height didn't really work correctly because the error screens (e.g. face not detected) required a classic navigation bar, causing various visual bugs (scrollbars, unexpected gradient, white-on-white back button). Now, this capture screen do need to integrate with the router and take the full height.

The solution is to remove the relative positioning on the `content` class, so the specific capture screen can be absolutely positioned relatively to the top of the frame. This is a risky change as it impacts *all* Web SDK screens. Luckily, the parent component is also relatively positioned thanks to the `step` class and this solution shouldn't change the layout in any way for other screens.

The navigation bar (back button, close icon on modal) needs to render transparently and white-on-black, so it is pre-rendered and passed to the component so it can display it on top thanks to a high-enough z-index. It's unusual, but better than duplicating the navigation logic and style in Motion components.

This change does not remove the technical debt that was previously introduced. It will be done in a follow-up commit.

Requires [this MR](https://gitlab.eu-west-1.mgmt.onfido.xyz/onfido/biometrics/experience/active-video-capture/-/merge_requests/73) on the `active-video-capture` package.

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "8.3.0",
       "license": "SEE LICENSE IN ./LICENSE",
       "dependencies": {
-        "@onfido/active-video-capture": "^0.7.0",
+        "@onfido/active-video-capture": "^0.8.0",
         "@sentry/browser": "^7.2.0",
         "blueimp-load-image": "~2.29.0",
         "classnames": "~2.2.5",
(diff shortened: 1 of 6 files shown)


victorcamargo-onfido:
The mentioned MR on `active-video-capture` is already merged. We'll release a new version.

PR #2057, opened 2022-08-23
feat(orc-1289): add new fields to profile data capture node

acMaiaOnf:
# Problem
New profile data capture node has configurable values and we want to show/hide data capture fields

# Solution

Set new data fields and required values based on the config from the profile capture node

## Checklist


src/components/Capture/PhoneNumberInput/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+  onChange: (ev: {
+    target: {
+      value: string
+    }
+  }) => void
+  smsNumberCountryCode?: Country
+  options: SdkOptions
+} & WithLocalisedProps
+
+const PhoneNumberInput = ({
```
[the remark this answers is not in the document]
acMaiaOnf: I tried that at first, but I was having more issues with that. I couldn't really get it work properly.
But we also don't need to duplicate it. I think I'll create profile data phone number input with the same lazy loading and styling, but with the new design.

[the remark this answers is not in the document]
josokinas: Or at least include whatever can be exported/imported from that original component, to reduce code duplication a little.

josokinas (approved):

src/locales/en_US/en_US.json
```diff
@@ -16,11 +16,18 @@
             "last_name": "Last name",
```
osolliec, line 16: could you please add strings for all languages at the same time please ?

osolliec (changes requested):

src/locales/en_US/en_US.json
acMaiaOnf: I don't have access to lokalise and I don't have the all the strings. I already asked for them, so I'll add them as soon as I can.

osolliec: there's a unit test in the dev branch that will fail if the locales are not in sync with the en_US file, just fyi :)

josokinas (approved):

josokinas:
One thing to note, that probably this would need to still support old way of providing SSN configuration, for legacy workflows.

acMaiaOnf:
It does @josokinas 
Currently it's passing the `ssn_enabled` flag with

`
      ssnEnabled={
        dataStep?.options?.ssn_enabled ||
        dataStep?.options?.profile_data_selection?.ssn_enabled
      }
`

acMaiaOnf:
And also for the other screens it's checking if it has the enabled flag or it's `true` or `false`

osolliec (approved):
would it be possible to minimize the diffs from the locales.json files ? just to ease the reviews regarding translations (maybe we should auto format them in prettierrc file, not sure).

PR #2088, opened 2022-09-06
Post release/merge 9.0.0 into development

DannyvanderJagt:
Merge 9.0.0 back into development


DannyvanderJagt:
@warrenseine @osolliec There were merge conflicts in the locales files (caused by re-order of keys) and in `components/ActiveVideo/Uploading/index.tsx` (caused by arranging of code?)

In both cases, I used the latest `development` branch as origin point as the differences where too hard to spot. Can you take a quick look if those files look okay?

osolliec:
> @warrenseine @osolliec There were merge conflicts in the locales files (caused by re-order of keys) and in `components/ActiveVideo/Uploading/index.tsx` (caused by arranging of code?)
> 
> In both cases, I used the latest `development` branch as origin point as the differences where too hard to spot. Can you take a quick look if those files look okay?

As far as i'm concerned, you can just take development locales and ignore the rest (They were all updated directly from lokalise - and the source of truth should be in lokalise). The only exception is the auth_* keys that are not in lokalise for some reason.

release/validateInput.js
```diff
@@ -60,5 +60,5 @@ const validateReleaseVersionNPM = async () => {
 
 ;(async () => {
   await validateInput()
-  await validateReleaseVersionNPM()
+  // await validateReleaseVersionNPM()
```
warrenseine, line 63: Is this commented code expected?

DannyvanderJagt: Yes, it blocked the release.

PR #2089, opened 2022-09-06
Move sdk configuration service into it's own module and use redux store

DannyvanderJagt:
# Problem
The SdkConfigurationService is a React Context and can't be used outside the React realm. We need to use it at different places, like the Logger etc. 

# Solution
- Put SdkConfigurationService into it's own module to clean up a bit
- Let the SdkConfigurationService use the Redux store instead of a local useState

The overall functionality stayed the same, only the place where we store the data is changed.

# Later
- When WebSDK Master plan architecture is decided, cleanup this module and make it independent from other parts (and files) of the SDK

## Checklist


src/components/ReduxAppWrapper/index.tsx
```diff
@@ -5,7 +5,7 @@ import { Provider as ReduxProvider } from 'react-redux'
 import reducer from './store/reducers'
 
 import type { CombinedActions, RootState } from '~types/redux'
-
+console.log('LOAD REDUX')
```
zoeradkani, line 8: is there reason we keeping the console.log?

src/components/ReduxAppWrapper/index.tsx
```diff
@@ -15,6 +15,8 @@ export const reduxStore = createStore(
     : undefined
 )
 
+console.log('reduxStore', reduxStore)
```
zoeradkani, line 18: should we remove this?

osolliec:
If you need specific configuration inside the logger, can't you request that configuration in the "constructor" of the logger ? Then only build the logger after you get the SDK configuration from SdkConfigurationServiceProvider.

The only downside is you can't use the logger until after you have the SdkConfiguration, but I think we get the configuration at the very beginning of the flow.

I'm not sure using redux is great unless we absolutely need a store - imho avoid it if possible

DannyvanderJagt:
This isn't a prefect solution indeed. The logger is a "new" category of internal interface as it has to be available everywhere and not only inside the React realm.

Regarding the proposed solution, we will need access to the SdkConfiguration config from outside the react realm in multiple places in the near future. Therefore duplicating requests doesn't seem like a good solution.

PR #2100, opened 2022-09-12
Feature/Integrate Passive Signals module

ethan605:
# Problem

As with the 1st iteration of Passive Signals module, we want to integrate it into the Web SDK and release it to our initial customers.

# Solution

Passive Signals module shall be loaded dynamically from a remote host, so there won't be any Node.js module installed.

Tasks to be done:

- [x] Integrate with a remote configuration to be able to turn off from our server
- [x] Start capturing at the start of the flow and finish once it ends

## Checklist

### DIFF
--- a/build/webpack/constants.ts
+++ b/build/webpack/constants.ts
@@ -78,10 +78,8 @@ const STAGING_CONFIG: ConstantMap = {
   USER_CONSENT_URL: 'https://assets.onfido.com/consent/user_consent.html',
   COUNTRY_FLAGS_SRC: 'https://assets.onfido.com/flags/',
   RESTRICTED_XDEVICE_FEATURE_ENABLED: false,
-
-  // FIXME: change this to the pre-prod URL when it's properly deployed.
   PASSIVE_SIGNALS_URL:
-    'https://assets.onfido.com/passive-signals/v1/signal.min.js',
+    'https://assets.onfido.com/dev/passive-signals/v1/signal.min.js',
 
   // @TODO: clean-up this config when v4 APIs are live
   USE_V4_APIS_FOR_DOC_VIDEO: process.env.USE_V4_APIS_FOR_DOC_VIDEO,
--- a/build/webpack/constants.ts
+++ b/build/webpack/constants.ts
@@ -25,9 +25,6 @@ export const PRODUCTION_BUILD = NODE_ENV !== 'development'
 export const SDK_TOKEN_FACTORY_SECRET =
   process.env.SDK_TOKEN_FACTORY_SECRET || 'NA'
 export const SDK_ENV = process.env.SDK_ENV || 'idv'
-// FIXME: change url to the one, where it's deployed live
-const PASSIVE_SIGNALS_URL =
-  'https://onfido-assets-production.s3.eu-west-1.amazonaws.com/dev/passive-signals/v1/signal.js'
 
 type ConstantMap = Record<string, string | boolean | undefined>
 
@@ -45,7 +42,8 @@ const PROD_CONFIG: ConstantMap = {
   USER_CONSENT_URL: 'https://assets.onfido.com/consent/user_consent.html',
   COUNTRY_FLAGS_SRC: 'https://assets.onfido.com/flags/',
   RESTRICTED_XDEVICE_FEATURE_ENABLED: true,
-  PASSIVE_SIGNALS_URL,
+  PASSIVE_SIGNALS_URL:
+    'https://assets.onfido.com/passive-signals/v1/signal.min.js',
 }
 
 const TEST_DEPLOYMENT_CONFIG: ConstantMap = {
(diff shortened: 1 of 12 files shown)


DannyvanderJagt:
Just a note: Maybe worth considering separating concerns and putting the Passive Signal into it's own module (or place like in `/core`). Right now it's more spread around through the code base and hard to write (unit?) tests for.

build/webpack/constants.ts
```diff
@@ -42,6 +42,8 @@ const PROD_CONFIG: ConstantMap = {
   USER_CONSENT_URL: 'https://assets.onfido.com/consent/user_consent.html',
   COUNTRY_FLAGS_SRC: 'https://assets.onfido.com/flags/',
   RESTRICTED_XDEVICE_FEATURE_ENABLED: true,
+  PASSIVE_SIGNALS_URL:
```
DannyvanderJagt, line 45: The file doesn't work at the moment

src/components/Router/HistoryRouter.tsx
```diff
@@ -45,7 +50,20 @@ export const HistoryRouterWrapper = ({
     )
   }
 
-  return <HistoryRouter {...props} {...useStepsProps} steps={steps} />
+  return (
+    <SdkConfigurationServiceContext.Consumer>
```
DannyvanderJagt, line 54: Any reason you're using a consumer instead of the hook inside the `HistoryRouter` itself?

src/components/Router/HistoryRouter.tsx
```diff
@@ -87,6 +106,33 @@ export const HistoryRouter = (props: HistoryRouterProps) => {
     })
   }
 
+  const [passiveSignalsTracker, setPassiveSignalsTracker] = useState<
```
DannyvanderJagt, line 109: I think it would be better to abstract the code into it's own hook, like `usePassiveSignalsTracker()`. Now the code it polluting the router.

DannyvanderJagt (changes requested):
- Overall I would suggest to move the PassiveSignal code into it's own place.
- The remote script (aws) can't be loaded
- Are there any cookies involved?

src/components/Router/HistoryRouter.tsx
```diff
@@ -87,6 +106,33 @@ export const HistoryRouter = (props: HistoryRouterProps) => {
...
+    PassiveSignals.Tracker | undefined
+  >(undefined)
+
+  useEffect(() => {
```
nulrich, line 113: Maybe the Router is not the best place to start the tracker. Can't the `PassiveSignalsTracker` could wait for the sdkConfiguration using redux?

src/components/Router/HistoryRouter.tsx
ethan605: Right, I forgot about that 🤦

src/components/Router/HistoryRouter.tsx
ethan605: Make sense, combining with the above suggestion, we can encapsulate the consumption of `useSdkConfigurationService` from `HistoryRouter`

src/components/Router/HistoryRouter.tsx
ethan605: Reverted as we don't need it anymore with `usePassiveSignals()` custom hook

src/components/Router/HistoryRouter.tsx
ethan605: Checked out as `usePassiveSignals` custom hook

ethan605:
> Just a note: Maybe worth considering separating concerns and putting the Passive Signal into it's own module (or place like in `/core`). Right now it's more spread around through the code base and hard to write (unit?) tests for.

Thanks for the suggestion. We're still in the middle of figuring out what is the best approach for this module to be integrated into Web SDK. Expecting major changes in the next versions (which will happen soon). At the moment, our top priority is to push this to our first few clients. Let's bring this into consideration in the next iterations :pray:

src/contexts/usePassiveSignals.ts
```diff
@@ -0,0 +1,39 @@
...
+    // Don't load the Passive Signals module by default
+    const enabled =
+      sdkConfiguration?.device_intelligence?.passive_signals?.enabled || false
+
+    if (!enabled || !token || !PASSIVE_SIGNALS_URL) {
+      return
+    }
+
+    loadExternalScript(PASSIVE_SIGNALS_URL, () => {
+      if (window.PassiveSignalTracker) {
```
warrenseine, line 25: If we leak something on the window, can we prefix it with Onfido?

src/components/utils/dynamicLoading.ts
```diff
@@ -0,0 +1,12 @@
+type OnloadCallback = () => void
+
+export const loadExternalScript = (src: string, onload: OnloadCallback) => {
```
warrenseine, line 3: We're already using [loading lazy components](https://github.com/onfido/onfido-sdk-ui/blob/development/src/components/ActiveVideo/Lazy.tsx) in the Web SDK using dynamic imports. Have you considered this option rather than using the good old script injection?

src/components/Router/HistoryRouter.tsx
nulrich: I'm resolving this, we'll do it later when @DannyvanderJagt is done with its PR.

src/contexts/usePassiveSignals.ts
ethan605: Good point, the name also needs some extra polishment.

src/components/utils/dynamicLoading.ts
ethan605: Thanks, I was fully aware of our lazy loading stuff. The reason for not going this way was it's not about lazy loading. It's about entirely control of the release cycles. As a separate developed module, we don't fancy being stuck with the Web SDK releases. Furthermore, there's already a plan to offer this module outside of the SDK world, so we need to keep it like this for the sake of flexibility.

DannyvanderJagt (approved):
Providing we figure out the stuff we mentioned on slack, it looks good to me 👍

src/contexts/usePassiveSignals.ts
ethan605: Changed to `OnfidoPassiveSignals`.

src/components/utils/dynamicLoading.ts
warrenseine: Ok, yes, that makes sense considering the RFC. Thanks!

warrenseine (approved):

